PR #8, opened 2020-07-23
Add eslint to the precommit hook for lint-fixed

brianespinosa:
This PR adds support for eslint and babel as portable package configs. Also adds lint to the build pipeline, and to the pre-commit hook for lint-staged.

Will eventually want to see if we can set up Rollup this same way... and possibly Cosmos.

## New Component Checklist


brianespinosa:
image

PR #13, opened 2020-07-27
DEV-11892 - integrate pagination component into FDS

lordpixel:
# Espressive Pagination
This PR integrates pagination to the FDS, it also fixes semmantic ui for cosmos. It contains the latest fixes applied to the original component.

## With a table
image

## With a list
image

## New Component Checklist

✅  Includes unit tests for _ALL_ required FDS use cases
✅  Does not mute any top level API props in React
✅  Includes an explanation for any changes to Jest snapshots (should be a PR tag automatically added if this happens)
✅   Component uses `React.forwardRef()` when a single DOM node is returned, _AND_ it makes React-sense to access that 
DOM node


brianespinosa:
@lordpixel it looks like this might be because your branch was out of date with the base. Github has a button to update that and requires it to be up to date before merge, which is great. That is sometimes our issue with inconsistency in Bitbucket but there is no way to enforce that there. 👍 
image

I am updating your branch with the base and that should start builds again. I suspect it will not fail lint this time.

lordpixel:
Thank you @brianespinosa!

design/ui/index.js
```diff
@@ -1 +1,2 @@
-export {EspButton} from './esp-button/dist';
```
brianespinosa, line 1: @lordpixel I am glad you saw this. I am not sure if this file will be manually updated in the long term... I suspect we can auto generate it at some point if it is needed.

This file is consumed by the FramerX tooling that generates all of the code components that Framer will use. Probably makes sense for us to manually add these exports to this file for now.

design/ui/pagination/README.mdx
```diff
@@ -0,0 +1,170 @@
+---
```
brianespinosa, line 1: @lordpixel none of the comments on this file are things for you to change... they are merely me sharing some thoughts about how this will eventually work in terms of the MDX tooling so we have a more shared understanding of the direction at a technical level. I think that will be helpful in terms of what sorts of things we might need to come back and clean up as the tooling improves.

design/ui/pagination/README.mdx
```diff
@@ -0,0 +1,170 @@
...
+name: Pagination
+---
```
brianespinosa, line 3: We will eventually have much more in the header schema of these documents... including the component name. I think we can make a lot of the MDX generation automatic from that with a template. Things like importing the component based on the name, adding it automatically to the Props component, etc.

design/ui/pagination/README.mdx
```diff
@@ -0,0 +1,170 @@
...
+
+import semanticUiCssOffline from 'semantic-ui-css-offline';
+import { Playground, Props } from 'docz';
```
brianespinosa, line 6: We will probably be building our own Props component that better defines what the props are. I think it will also exist in a drawer on the right side of the docs so you can see the props at all times while using a Playground. The Docz Props component also does not tell us what the enum values are, which feels like a big missing piece that we could add with our own. 

We will also probably build our own Playground component with react-live. My hope is that the playground will auto generate all of the playgrounds from a Cosmos fixture named export. So if we have 6 fixtures exported from a single file, our Playground component would create all of them in our doc too. 

Both of the above custom components for the docs would in theory auto import the component and fixture based on the component name defined in the head. Or you could manually define the component too if you needed to.

The reasoning behind this investment is I think we hope that also UX designers will help with this documentation. If we can make it simpler to create this content, and keep it consistent, that is going to pay off for us.

design/ui/pagination/README.mdx
lordpixel: got it!

design/ui/pagination/README.mdx
```diff
@@ -0,0 +1,170 @@
...
+## Props
+| Name    | Default    | Type | Description |
+| --------|------------|------| ------------|
+|entityNamePlural|`Null`|String|Entitiy's name in plural form|
+|entityNameSingular|`Total`|String|Entitiy's name in singular form|
+|isLoading|`false`|Boolean|Specifies if the data is loading|
+|onPaginationChange|`noop`|Function|Upstream logic to be called on change|
+|limit|`10`|Number|Number of items to display per page|
+|page|`1`|Number|Current page number|
+|totalItems|`0`|Number|Total count of items to be paginated|
```
brianespinosa, line 25: I think leave this table for now... but essentially, this table should be generated automatically for us from static analysis of the component using the Props component.

The one thing I think we lose though... is that this spec forced us to make decisions about code and how the component would work before we started writing code. So, we need to figure out how we want to define spec moving forward

For this project since it is centered around UX as much as code, we should have a spec definition that also encompasses those ideas as well. That spec pattern only exists as an idea right now. I think it makes sense for that spec to also live in this repo somewhere. Possibly as a separate MD file that we can write things on, and one that does not change as we change the component code so we can A/B against it to see if we are still following what we wanted.

If we did have a spec MD file, we could actually locate it in the package folder. And it could be a separate tab on the doc page for a component. It could also potentially be used to track the progress of the components we are working on and what our needs are. @CaffeinatedAllie that could actually be what we use to generate a table of components and patterns, and their progress.

design/ui/pagination/README.mdx
```diff
@@ -0,0 +1,170 @@
...
+<Playground>
+  <Pagination
+    entityNamePlural={'Employees'}
+    entityNameSingular={'Employee'}
+    isLoading={false}
+    limit={10}
+    page={1}
+    count={100}
+  />
+</Playground>
```
brianespinosa, line 170: @lordpixel curious how you feel as a developer about what these two components generate and if you feel it is useful enough as is or not.

design/ui/pagination/README.mdx
brianespinosa: @CaffeinatedAllie these comments are also probably useful for you as well, since these docs, curating them, and maintaining them, need to serve both developers and designers.

design/ui/pagination/cosmos.decorator.js
```diff
@@ -0,0 +1,11 @@
+// The only thing this decorator will do for now is import our SUI custom CSS.
+// There are currently issues with the out of the box compiled styles there
+// and possibly some missing loaders so we will use a different variant for now
+// that does not have any missing font/image dependencies.
+
+// import 'semantic-ui-css-custom';
```
brianespinosa, line 6: @lordpixel one the package you are working on is updated, using that custom CSS package here in this decorator should give us what we need. Note that the normalize CSS I THINK is included in that custom CSS. If not, we should do the normalize.css import above the custom SUI. If normalize is included in that custom SUI file then we can remove it here.

I think this should probably be included as a global cosmos decorator too for all instances of Cosmos being run. That way we do not have to define it again and again... and we can also see where/if there are any class collisions as we move forward to our own custom stuff.

brianespinosa: Again... no change needed. Just informational.

design/ui/pagination/package.json
```diff
@@ -0,0 +1,45 @@
...
+  "name": "@espressive/pagination",
+  "version": "1.0.0",
+  "main": "dist/index.js",
+  "module": "es/index.js",
+  "scripts": {
+    "rollup": "rm -rf dist && rm -rf es && BABEL_ENV=production rollup -c",
+    "rollup:watch": "rm -rf dist && rm -rf es && BABEL_ENV=production rollup -c --watch"
+  },
+  "eslintConfig": {
+    "extends": "react-app"
```
brianespinosa, line 11: These will need to get changed this week to 'espressive' instead of 'react-app' now that we have our own. 🙌

design/ui/pagination/package.json
```diff
@@ -0,0 +1,45 @@
+{
+  "name": "@espressive/pagination",
+  "version": "1.0.0",
+  "main": "dist/index.js",
+  "module": "es/index.js",
+  "scripts": {
+    "rollup": "rm -rf dist && rm -rf es && BABEL_ENV=production rollup -c",
+    "rollup:watch": "rm -rf dist && rm -rf es && BABEL_ENV=production rollup -c --watch"
```
brianespinosa, line 8: These commands need to become make commands to make them more portable and easier to add to packages. But... they MIGHT also become something we run at the root of the package if this is a thing we need to do across multiple packages consistently. I still do not know what the DX is going to be for us when working on these. For now, they can stay scoped to the packages alone.

design/ui/pagination/rollup.config.js
```diff
@@ -0,0 +1,41 @@
...
+      use: ['sass'],
+    }),
+    babel({
+      babelrc: false,
+      babelHelpers: 'runtime',
+      presets: ['espressive'],
+      plugins: ['@babel/plugin-transform-runtime'],
+    }),
+  ],
+};
```
brianespinosa, line 41: Another file that needs to become an espressive package so we do not have to copy paste it everywhere. 

@lordpixel none of these things so far are anything you need to change. Just sharing vision on how some of this will work now that you have more context. I am marking things as viewed.

I will mark changes separately (if any)

design/ui/pagination/src/Pagination.js
```diff
@@ -0,0 +1,292 @@
...
+  /** entity name in singular form */
+  entityNameSingular: pt.string,
+  /** specifies if the data is currently loading */
+  isLoading: pt.bool,
+  /** items per page limit */
+  limit: pt.number,
+  /** a message to be displayed when no items to display */
+  notFoundMessage: pt.string,
+  /** current page */
+  page: pt.number,
```
brianespinosa, line 28: Semantics here also. I think this makes more sense to call this `currentPage` so it is easy to understand when consuming and viewing these components in code. What do you think?

design/ui/pagination/src/Pagination.js
```diff
@@ -0,0 +1,292 @@
...
+  /** the items collection length */
+  count: pt.number,
+  /** entity name in plural form */
+  entityNamePlural: pt.string,
+  /** entity name in singular form */
+  entityNameSingular: pt.string,
+  /** specifies if the data is currently loading */
+  isLoading: pt.bool,
+  /** items per page limit */
+  limit: pt.number,
```
brianespinosa, line 24: `limit` also feels like I do not know what it is for without looking at the note about being for the page.

design/ui/pagination/src/Pagination.js
```diff
@@ -0,0 +1,292 @@
...
+
+const propTypes = {
+  /** Can render as a different tag or component */
+  as: pt.oneOfType([pt.string, pt.node]),
+  /** the items collection length */
+  count: pt.number,
+  /** entity name in plural form */
+  entityNamePlural: pt.string,
+  /** entity name in singular form */
+  entityNameSingular: pt.string,
```
brianespinosa, line 20: I wonder if there is a way for us to reuse the plurals logic in our translations and duplicate that here. That way we can just pass a single entity object from our translations and have it work without us having to separate it out. It would also reduce the API surface area.

What are your thoughts on this? If we went that route, maybe the object signature would be 1:1 the same as how we define a plural in POEditor? That would also give better localization support since some languages the plural changes based on the count, like Russian has words where it has singular, plural 2-9, plural 10+

design/ui/pagination/src/Pagination.module.scss
```diff
@@ -0,0 +1,87 @@
+._ {
+  display: grid;
+  grid-template-columns: 50px 1fr 50px;
+  grid-template-rows: 50px 1fr 50px;
+  gap: 0px 0px;
+  grid-template-areas: "hback header hfrwd" "content content content" "fback footer ffrwd";
```
brianespinosa, line 6: @lordpixel it breaks my heart... but until we are no longer supporting ie11... I think we might not be able to use CSS grid. 💔  It requires vendor prefixes to work there and does not fully support the whole spec. Are these CSS files the same as what that component is using in our production app?

PostCSS might be saving us there by adding those prefixes for us.

design/ui/pagination/src/Pagination.test.js
```diff
@@ -0,0 +1,93 @@
+import React from 'react';
+import { cleanup, fireEvent, render } from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
```
brianespinosa, line 3: create-react-app does a good job of always importing this extend-expect. It would be nice if we also had that in the future.

design/ui/index.js
brianespinosa: I am going through and resolving all conversations that are NOT part of requested updates.

design/ui/pagination/src/Pagination.js
```diff
@@ -0,0 +1,292 @@
...
+import 'semantic-ui-css-custom';
+import styles from './Pagination.module.scss';
+
+const cx = classNames.bind(styles);
+
+const propTypes = {
+  /** Can render as a different tag or component */
+  as: pt.oneOfType([pt.string, pt.node]),
+  /** the items collection length */
+  count: pt.number,
```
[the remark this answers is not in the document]
lordpixel: you are right, recordCount is more semantic. Leme change it.

design/ui/pagination/src/Pagination.js
lordpixel: absolutely, working on it.

design/ui/pagination/src/Pagination.js
lordpixel: I am pretty sure we can replicate the behavior you mention for this. Also, I was thinking on translations but since I don't know if that's already set up I didin't explore that path.

Let me explore it and share my findings.

lordpixel:
> Looks like only the prop naming on a few might need to change to be more semantic. This is something we have been very bad at forcing people to think about in our main application, but for a consumable library, it is VERY important.
> 
> I think leaving the CSS grid stuff is probably okay for now. I am not sure if it will live on long term. But if it does, there will likely be grid templates at the library level which we can reach for. Lots of styles will need to be cleaned up so we can also make the CSS more dry and define consistent elements with design tokens and variables.

how do we mark this as resolved?

brianespinosa:
@lordpixel this PR is still failing install. Take a look at the details for that build step and see if you can figure out what is blocking. If you are unable to figure it out, please @ me in Slack and we can look together.

The links from the checks should take us directly to the build steps and the console output of the failure. It is WAY easier to track this stuff down in Circle. 🙌 
image

lordpixel:
> @lordpixel this PR is still failing install. Take a look at the details for that build step and see if you can figure out what is blocking. If you are unable to figure it out, please @ me in Slack and we can look together.
> 
> The links from the checks should take us directly to the build steps and the console output of the failure. It is WAY easier to track this stuff down in Circle. 🙌
> image

@brianespinosa I've been using Circle since you recommended me to do so, I'm loving it! But, should this command include **pagination** as well?

design/ui/pagination/src/Pagination.js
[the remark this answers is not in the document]
lordpixel: @brianespinosa the reasoning behind renaming some props was precisely to be consistent with BE as they use count. My plan was to update the implementation once we switch to FDS, but you are right on the fact that we should provide a major version bump in the older commponent. I will work on this and keep you posted.

brianespinosa:
@lordpixel are you planning on doing the Makefile work as part of this PR? I am realizing that the docs deploy process I think may also fail due to the changed command. I think I need to introduce a netlify config so we can control the deploy commands from repo configuration in code versus in the Netlify UI. So you may need that to add to this too. I will be adding that today while working on Docz so you'll have access to that in develop in the next hour or two.

lordpixel:
> @lordpixel are you planning on doing the Makefile work as part of this PR? I am realizing that the docs deploy process I think may also fail due to the changed command. I think I need to introduce a netlify config so we can control the deploy commands from repo configuration in code versus in the Netlify UI. So you may need that to add to this too. I will be adding that today while working on Docz so you'll have access to that in develop in the next hour or two.

@brianespinosa for the _Makefile_ work I'd like to branch out from develop and merge that one first, that way we keep it atomic, also, this PR is big already and can wait.

Regarding Netlify config, a pull in develop didn't bring any updates, I'll keep an eye on it

brianespinosa:
> Regarding Netlify config, a pull in develop didn't bring any updates, I'll keep an eye on it

@lordpixel pull develop now. The `netlify.toml` file is now present.

design/ui/pagination/src/Pagination.test.js
```diff
@@ -0,0 +1,102 @@
...
```
brianespinosa, line 3: @lordpixel FYI we do not have to import extend expect anymore. It is part of our Jest config. Auto imports it for us. Just FYI. No need to change here for now.

design/ui/pagination/src/Pagination.test.js
```diff
@@ -0,0 +1,102 @@
...
+
+/**
+ * We import the fixture to save us from the setup hassle,
+ * the fixture contains everything we need for testing
+ * the component */
+import { Wrapper } from './Pagination.fixture';
+
+describe('Espressive UI Pagination', () => {
+  afterEach(cleanup);
```
brianespinosa, line 12: We also do not have to do `afterEach(cleanup)` anymore in our Jest tests. We have it configured to automatically do this now. 🎉 No need to change here for now either.

brianespinosa:
@lordpixel this LGTM once the conflicts are resolved. Looks like merging the makefile branch first introduced them.

PR #14, opened 2020-07-27
DEV-11910 - fix semantic ui for cosmos

lordpixel:
## New Component Checklist


brianespinosa:
@lordpixel we definitely do not want to check all of SUI in to VC here. Technically... we do not actually want any of the CSS tooling as part of this repo at all. I think the offline CSS is only to help us check components that might be using SUI and need styles. But in the long run, we are going to have everything custom.

Could you add some explanation of what this PR is adding? Is the goal simply to get some CSS that will work for our use case for now?

lordpixel:
We are closing this PR in favor of another one without all these assets.

PR #15, opened 2020-07-27
fix docz

brianespinosa:
Removing checklist since this does not impact components.

This fixes Rollbar output so it no longer compiles runtimes that have prop-types stripped. In the long run... trying to use a config that works the same as react-scripts for the runtimes in Rollbar might not actually be the best path forward since our needs are different as a package library. We might need to have our own config that we maintain ourselves for building runtimes of components. This will mean we also needs some tests on making sure these components also work in other environments on their own. Lots of testing work to be done on the runtimes themselves separate from the unit testing.

Once this PR gets merged and we make another merge to main from develop... I think this will also deploy correctly on Netlify.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -70,7 +70,7 @@ jobs:
                 at: ~/
             - run:
                 name: Build
-                command: yarn test
+                command: yarn workspace button run rollup && yarn docz:build
             - slack/status:
                 fail_only: true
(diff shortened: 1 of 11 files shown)


brianespinosa:
Love that you can see who reviewed and when in the timeline too. 
image

brianespinosa:
@lordpixel this is okay to review now. No more pushes coming with builds. config updates worked.

PR #20, opened 2020-07-29
Move robots.txt to docs build root

brianespinosa:


brianespinosa:
Verified with build preview that robots.txt is now at root.

PR #21, opened 2020-07-31
Test new Netlify.toml config

brianespinosa:
@lordpixel this is the file you need. Should already be present in develop if you pull that into your branch.


brianespinosa:
Deleted all existing Netlify build settings just to make sure that the `netlify.toml` file is our source of truth.

PR #22, opened 2020-07-31
rollup refactor and docz spec

brianespinosa:
No new components.

- Refactors Rollup configuration to generate a package relative `rollup.config.js` file which pulls all Rollup, Babel, and PostCSS options also from the repo root.
- Cleans up package script commands now that they can be more simple
- Relocates package dependencies for Rollup, Babel, PostCSS, SASS to the monorepo root now that they do not have to be package relative
- Adds a separate env for Jest tests, since the babel.config.js file in the root was previously only used by Jest.

- Adds some additional documentation tests to see how we might be able to define spec files separate from monorepo packages (which works :partyparrot:)

NOTE: Now that this package technically is using its own custom Babel config for building runtimes... we may want to explore officially moving the `/configs` stuff officially out to their own repo and finally publishing some packages and make them available in `app_web` too. The babel config is not being used here. But the eslint rules certainly are. The eslint rules do still need to get extended.


babel.config.js
```diff
@@ -1,24 +1,25 @@
...
+      '@babel/preset-env',
       {
+        modules: false,
         loose: true,
       },
     ],
+    '@babel/preset-react',
+  ],
+  plugins: [
+    '@babel/plugin-proposal-optional-chaining',
```
brianespinosa, line 13: @Espressive/front-end This is something that would be great to have in all of our applications to be honest. https://github.com/tc39/proposal-optional-chaining

brianespinosa:
Forgot to also update the build command in the circleci.yml file. We do need to unify these commands to Make sometime in the next couple weeks.

PR #24, opened 2020-08-04
Button Tests & lots more eslint rules

brianespinosa:
- Adds button tests and sticks to some best practices
- Defines some lint rules for tests to keep some things consistent
- Adds a lot more eslint rules across the board which have auto fix support... much faster DX
- Cleans up any tiny issues from lint rules added
- Few improvements to our precommit hook for lint/prettier


brianespinosa:
`Button.test.snap` changed in this PR. So we SHOULD be getting a label added to the PR telling us this. Seems like the labeler addon is not working correctly. Will keep this PR open to try to trigger that again and get the label working.

doczrc.js
```diff
@@ -17,5 +13,9 @@ export default {
         },
       },
     },
+    initialColorMode: 'dark',
```
lordpixel, line 16: 🙌  dark mode FTW!

PR #44, opened 2020-08-05
add Makefile with common commands

lordpixel:
image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,17 @@
+RM=rm -rf
+
+.PHONY: all clean
+
+all:
+	yarn install
+	yarn workspace button build
+
+watch:
+	yarn workspace $(P) watch
+
+build:
+	yarn workspace $(P) build
+
+clean:
+	$(RM) yarn.lock node_modules
+	yarn workspace button clean
--- a/TODO.md
+++ b/TODO.md
@@ -57,8 +57,8 @@
 
 ## Make
 
-- [ ] Set up common commands and make recipes area to simplify tooling maintenance and process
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -0,0 +1,17 @@
...
+	yarn workspace button build
+
+watch:
+	yarn workspace $(P) watch
+
+build:
+	yarn workspace $(P) build
+
+clean:
+	$(RM) yarn.lock node_modules
```
brianespinosa, line 16: We probably do not want the lockfile and node_modules folders in a typical clean command that gets run. This might be better to call this something like a `force-clean` or something like that. This way we know it is a higher level command than a typical clean.

We will definitely want a clean command that does things like cleanup on runtimes from packages as well as build folders, which should not touch lockfile or modules folders.

Makefile
```diff
@@ -0,0 +1,17 @@
...
+
+all:
+	yarn install
+	yarn workspace button build
+
+watch:
+	yarn workspace $(P) watch
+
+build:
+	yarn workspace $(P) build
```
brianespinosa, line 13: Perfect. <3

Makefile
```diff
@@ -0,0 +1,17 @@
...
+
+clean:
+	$(RM) yarn.lock node_modules
+	yarn workspace button clean
```
brianespinosa, line 17: If you go make a change to this clean command, I think also quickly add a `help` command at the top of the file and also set that as the default command for the makefile.

Makefile
brianespinosa: @lordpixel nevermind. I am resolving these issues. We can clean those things up separately now that this file is here.

PR #49, opened 2020-08-06
Resolves #47 repository structure updates

### LINKED ISSUE #47: Minor updates to repository structure preparing for package publishing
Initially when the repo was set up we thought this code would potentially be a POC and eventually be dropped into the main `app_web` monorepo. Seeing all of the tooling involved here, and realizing that having a different package also helps separate concerns... we are going to keep this as a separate repository from all of our implementation details.

Based on this, the current structure needs to get updated so that things are easier to maintain in terms of our revised use case.

- [ ] the system is the focus of the repo now
- [ ] single entry point into the package to make consuming and building easier

brianespinosa:
Ends up that this was a little more than a minor directory update. But makes things way more consistent for moving everything forward.

Biggest things to see:
- Moves all source assets to three folders depending on how they are intended to be exported:
  - framer
  - src
  - utils
- Moves some previous, complex root level config into their own dirs:
  - .cosmos
  - .jest
- Deprecated use of running cosmos scoped to a specific package to keep toolchain maintenance easier
- Moved some complex commands for test builds on Cosmos into root package.json for easier reuse


brianespinosa:
@CaffeinatedAllie this also means `specs` and `tokens` are now located in the src folder. But at least this makes things more consistent where we look for everything.

brianespinosa:
New: 
image

Old: 
image

CaffeinatedAllie:
TBH I would look for the spec files in the SRC folders anyways :) So I approve of this change! 👍

brianespinosa:
:partyparrot:

PR #52, opened 2020-08-06
Do another deploy 🙌🏽

brianespinosa:


brianespinosa:
We can leave this for now... but I think we will not be worrying about having another deploy until the docs are being build from next instead of docz

brianespinosa:
Needs #69

brianespinosa:
👀

brianespinosa:
Okay, deploys now working here. Will eventually need to add slack auth with lambda functions. For now, looks good.

Issue #35, opened 2020-08-04
Set up first Layout components and sub exports before we publish first MVP


brianespinosa:
Layouts are going to be an important concept of this FDS. We should make sure to publish at least one Layout with the package along with whatever component(s) come with it as well. Initially it seems like we will have two of these:

Barista
Admin

PR #54, opened 2020-08-07
#35 layouts and member exports

brianespinosa:
Adds a new type of component: LAYOUT!

## New Component Checklist


brianespinosa:
Need to remove prop API. Probably need to have these work with only children. Set all layout values to #root node. Hopefully having root CSS import there still works as expected for global styles if root level layout only returns fragment and child nodes.

brianespinosa:
Okay... this is close-ish for now. I discovered some issues with how we are defining imports of component groups and to make that easier we needed some more babel plugins.

There are also some significant tests for theming... both in terms of branding, checking contrast ratios for colors using a single branding variable, and also validating some things for dark mode. This also has some initial POCs for reusing common style patterns for things like styling scroll areas.

There is going to be almost as much work defining the tools to reach for there. But honestly... some of the reusable style attribute stuff is going to be tricky to do with any sort of CSS in JS. The risk is that people may start mutating styles that we want. And in our case for this library... we actually want to enforce some of that. Possibly even have some custom lint rules to help.

Dart sass has a number of interesting, new patterns. There are some gotchas to get used to from node-sass.

src/layout/admin/src/Admin.fixture.js
```diff
@@ -0,0 +1,130 @@
+import React from 'react';
+import Admin from './';
+
+const longContent = (
```
brianespinosa, line 4: These are kind of annoying. Considering maybe adding `faker` for doing these things.

PR #69, opened 2020-08-24
New docs!

brianespinosa:
There is probably a lot to look at here...

- Removez docz
- Fancy, new, custom Nextjs doc generation
  - Mostly POC
  - Should work with all existing MDX files
  - No "page level" support for MDX files yet
- Updates build commands for CI
- Updates netlify.toml for publishing


packages/cascara/README.md
```diff
@@ -0,0 +1,9 @@
+# @espressive/cascara
+
+> NOTE: This will eventually get replaced with a generated README which will probably link to the cascara.design docs.
+
+If you happen to be here ahead of schedule:
+
+```javascript
+import { Some, Cool, Components } from '@espressive/cascara';
```
lordpixel, line 8: 🙌

packages/cascara/src/ui/Button/Button.doc.mdx
```diff
@@ -21,10 +19,22 @@ Still to also explore here:
...
 <Playground>
   <Button content='Docz Demo' />
 </Playground>
 
-<Playground>
-  <Button content='Positive' outcome='positive' />
-</Playground>
+What if wrapping our code examples in regular markdown code blocks worked... but if we use the jsx syntax, we don't just show the code with syntax highlighting. Instead we show an example of the component in this state and a live editor.
+
+```jsx
```
lordpixel, line 30: You did it! awesome stuff here! 🖤

brianespinosa:
@lordpixel I am merging. Still a lot more to do on the docs, but I think, other than not having prop type documentation, we are almost as good as what docz is without using it. Will have to try to circle around before the end of this patch release to get those on there too. We do want documentation to also ship with out code. That should be non-negotiable if we are going to support developers.

The tricky part to see is if this actually works with deploying to netlify.... we will see in a few min.

PR #71, opened 2020-08-24
create table technical design

lordpixel:
This is a draft of the technical design for Table.
It's still in progress but I think it's good enough to start the discussion.


packages/cascara/src/ui/Table/TechnicalDesign.mdx
```diff
@@ -0,0 +1,138 @@
...
+  - Users
+  - Locations
+
+Caffeine
+
+- Configuration
+- Entities
+- Localization
+- Training
+- Variables
```
brianespinosa, line 28: This skips mention of the "status" app.

Probably also worth noting that views like this are also going to be needed in the new version of Doppio.

There are other applications that will be needed in the future as well. Might not need to mention that, but when we are taking inventory, we probably want to include all of our application buckets. @CaffeinatedAllie will be good at helping us with this more broad view.

packages/cascara/src/ui/Table/TechnicalDesign.mdx
```diff
@@ -0,0 +1,138 @@
...
+| callback  | Function | `noop`       | The function to be called when the action button is clicked |
+| trigger   | ANY      | `<Button />` | The element that will trigger the action                    |
+
+### Data
+
+The data passed to the table for rendering purposes.
+
+| Name | Type          | default | Description                                   |
+| ---- | ------------- | ------- | --------------------------------------------- |
+| data | Array[Object] | `[]`    | the tabular data to be displayed by the table |
```
brianespinosa, line 112: I wonder if there are some consistent things we could at least make sure are present in these data objects. I guess the only thing for sure that we would want to make sure of is `eid` which could then be used for keys in the table view. Past that, these objects could look like almost anything.

An `eid` will be super important for data arrays on anything virtualized too.

Not sure if we should necessarily address that here. 

Also, like the pattern of CRUD operations, I think that a `data` prop is also going to be some of our bread and butter in Cascara. We should also consider a section for making notes of how this particular prop should be used too.

packages/cascara/src/ui/Table/TechnicalDesign.mdx
```diff
@@ -0,0 +1,138 @@
...
+### Config
+
+There will be use cases where features like multi-selection, CRUD, custom actions, etc. won't be required at all, they must be disabled by thefault but easily activated. It makes sense to have boolean props that controll such features.
+
+Another important aspect of configuration is the definition of columns, hence a complete section is devoted to it.
+
+| Name    | Type          | default | Description                                        |
+| ------- | ------------- | ------- | -------------------------------------------------- |
+| Bulk    | Boolean       | `FALSE` | If true, activates the bulk mode (multi-select)    |
+| Columns | Array[Object] | `[]`    | Holds the definitions for each column in the table |
```
brianespinosa, line 91: When I got the the columns section below, I realized that if we are going to use consistent patterns in Cascara for how some of our props work... that might mean we want to name columns differently here. That way if I look at this prop as a developer, if I have seen and used it before on another Cascara component, I already know how it works here.

Maybe a better name for this is `dataConfig`. This would keep it sorted alphabetically with the `data` prop that it controls (and I love that from an OCD perspective), but that also would make it so we could use a `dataConfig` prop on any other Cascara components that use the same `data` > `dataConfig` > `doMagic` pattern. The objects expected in `dataConfig` could be consistent everywhere. If we are working with a List, Form, Table... if we see a `dataConfig` prop, we know that it is going to control how the data looks in the component we are applying it to, and it should have mostly the same object shape of the objects in the array.

I guess in theory... if we have a Table with a `data` and `dataConfig` defined, we might be able to swap it over to a List component instead, and except for a few minor tweaks to parameters on the `dataConfig` controlling where the data shows on the list items, it would immediately work.

packages/cascara/src/ui/Table/TechnicalDesign.mdx
```diff
@@ -0,0 +1,138 @@
...
+| ---- | ------------- | ------- | --------------------------------------------- |
+| data | Array[Object] | `[]`    | the tabular data to be displayed by the table |
+
+### Columns
+
+The definition of columns depends on the data, we must make sure we have a well-defined set of data types to allow display and mutation for each one.
+
+| Attribute  | Type                                                                   | default    | Description                                              |
+| ---------- | ---------------------------------------------------------------------- | ---------- | -------------------------------------------------------- |
+| accessor   | String                                                                 | `null`     | The name of the entity attribute that maps to the column |
```
brianespinosa, line 120: - [ ]  Thoughts on just calling this `attribute` instead? I am guessing that would be more simple

packages/cascara/src/ui/Table/TechnicalDesign.mdx
```diff
@@ -0,0 +1,138 @@
...
+| context    | Object                                                                 | `{}`       | the context to be passed down to the custom component    |
+| isEditable | Boolean                                                                | `FALSE`    | Specifies if the cells are editable                      |
+| label      | String                                                                 | `''`       | The text to display as column header                     |
```
brianespinosa, line 123: - [ ]  This LGTM... but perhaps could we make the description more generic? If we are going to reuse this object shape, what if the description was `Label text to display with this attribute`?

packages/cascara/src/ui/Table/TechnicalDesign.mdx
```diff
@@ -0,0 +1,138 @@
...
+| data | Array[Object] | `[]`    | the tabular data to be displayed by the table |
+
+### Columns
+
+The definition of columns depends on the data, we must make sure we have a well-defined set of data types to allow display and mutation for each one.
+
+| Attribute  | Type                                                                   | default    | Description                                              |
+| ---------- | ---------------------------------------------------------------------- | ---------- | -------------------------------------------------------- |
+| accessor   | String                                                                 | `null`     | The name of the entity attribute that maps to the column |
+| context    | Object                                                                 | `{}`       | the context to be passed down to the custom component    |
```
brianespinosa, line 121: Walk me through the use case with this. Is this actually React context?

packages/cascara/src/ui/Table/TechnicalDesign.mdx
```diff
@@ -0,0 +1,138 @@
...
+| isEditable | Boolean                                                                | `FALSE`    | Specifies if the cells are editable                      |
+| label      | String                                                                 | `''`       | The text to display as column header                     |
+| render     | JSX                                                                    | `null`     | A custom component to render                             |
+| style      | Object                                                                 | `{}`       | Custom style for the rendered component                  |
```
brianespinosa, line 125: I worry about exposing either `render` or `style` props on these. In theory, Cascara should be providing all of the display options needed for data. Essentially, if there is need for a new display style for something in a List, Table, Card... we would want to add that to Cascara and control that by the `type` attribute.

Opening up these two is probably too much and allows for the design system opinions to be bypassed. This is a slight shift in thinking from a UI library where we actually want to expose as many permutations for layout as possible. With Cascara we might be able to do these sorts of things with atomic pieces... but we might not be exporting all of those as public members. Cascara is the functional design system. So it has a lot more opinion added to it in terms of how things should work. Theoretically, that makes the APIs more simple that we expose... but the guts can be more complex than a UI library because of it.

- [ ] Discuss not using a render prop or exposing style

packages/cascara/src/ui/Table/TechnicalDesign.mdx
```diff
@@ -0,0 +1,138 @@
...
+| type       | Enum[`boolean`, `date`, `image`, `link`, `number`, `string`, `render`] | `'string'` | Specifies the type of data for the column                |
+
+### Data types
+
+The data types this table can handle are the following:
+
+For display puposes: `boolean`, `date`, `image`, `link`, `number`, `string` and `render`. The `render` type being a custom JSX.
```
brianespinosa, line 132: We should figure out how to identify which of these is part of MVP and which is future.

Also, we need to figure out what these little guys are called. They will be used in a reusable manner all over Cascara. Basically they are like display atoms or something. And they will have an editing experience as well.

- [ ] Maybe we just call these guys "atoms". And for all atoms they have an editing or display presentation?

packages/cascara/src/ui/Table/TechnicalDesign.mdx
brianespinosa: I guess if there was a future scenario where we wanted composability in this pattern, we could include render... but I don't think we should call it `render` since there is the render prop pattern in React and people may confuse it with that. If we do happen to include this flexibility at some point... maybe we call this `component` instead. and it would override what is going on with `type`.

If we did that, I think it would also be good if we do not actually expose the `component` prop as part of the docs. Maybe it is hidden just in case. Still I think that might not be an MVP thing.

packages/cascara/src/ui/Table/TechnicalDesign.mdx
```diff
@@ -0,0 +1,138 @@
...
+
+For mutation purposes, the implementation will support `select`, `input`, `boolean switch`, `date / time pickers`, `lookup select`. With the ability to expand to more types.
+
+## Context
+
+All the internal logic will be hosted in a Context. The definition is WIP.
```
brianespinosa, line 138: Okay, I can tell this is specific to React context. 👍🏽  Moving forward, any time we are referring to something that is part of React as part of the API, or any other tool as part of the API, we should figure out a consistent way of calling that out. We have a lot of terms in our ecosystem that have special meaning which we want to be clear on.

packages/cascara/src/ui/Table/TechnicalDesign.mdx
```diff
@@ -0,0 +1,138 @@
...
+Some tables allow the user to create a new entry in the table. Adding an empty row.
+
+**Inline edition**:
+Some tables allow the user to edit the data, replacing the display representation with some JSX that allows the mutation of the data.
+
+**Item deletion**:
+Some tables allow the user to delete record via a button displayed at the row level.
+
+**Custom actions**:
+In most implementations, tables add a trailing column to each row. This row can display `action buttons`, which can be mapped to corresponding actions over the entire row. i.e. edit, delete, cancel, etc.
```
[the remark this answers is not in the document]
lordpixel: completely agree, I will make the change

packages/cascara/src/ui/Table/TechnicalDesign.mdx
```diff
@@ -0,0 +1,138 @@
...
+
+## Data representation / mutation
+
+The way each cell represents the data for either display or mutation purposes not only depends on the data type but also on business logic, in most cases, values are easily coerced to string for display purposes and mutations are easily achieved via a controlled input component. The rest of scenarios require a more complex implementation, e.g. the `employee` field in the `Users` table from `Barista Admin` which under the carpet is a single `eid`, but the representation of that cell is an avatar with the user name it points to, although it could be any other representation, specially when editing the value.
```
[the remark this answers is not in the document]
lordpixel: by `mutation` I mean `to edit` the cells in a row and then notify upstream (parent component) about the change. I gues `edit` is a better term.

The header is misleading, it should read `Data representation for display and edit purposes`. The point I am trying to make is that the representation of the data depends on whether the cell is in`display` or `edit` mode.  

> The rest of scenarios require a more complex implementation, e.g. the `employee` field in the `Users` table from `Barista Admin` which under the carpet is a single `eid`, but the representation of that cell is an avatar with the user name it points to

packages/cascara/src/ui/Table/TechnicalDesign.mdx
```diff
@@ -0,0 +1,138 @@
...
+
+# Proposal
+
+Much thought has been put to the proposal described in this section, our goal is to create a table that can support all of our current use cases while at the same time being abstract enough to support future ones. We are not inventing the wheel here, there are great open source libraries like `react-table`. The problem with those libraries is that they are bloated with things we don't need either because we already abstracted such functionality out to other components or we sinply don't need it.
```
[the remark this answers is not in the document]
lordpixel: LOL the `m` in my keyboard seems to be broken, I probably hit the wrong key with the correct force 😜 , will fix thx.

packages/cascara/src/ui/Table/TechnicalDesign.mdx
lordpixel: it actually reads better, will fix, thx!

packages/cascara/src/ui/Table/TechnicalDesign.mdx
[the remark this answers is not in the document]
brianespinosa: Okay. This makes more sense then. 👍🏽 

This brings up another point that I feel this spec might not capture now that you mention this. We should probably capture somewhere a statement that the data does not get edited directly inside of the table Context or State itself. If data is in an edit mode, and a CRUD action fires, the logic for submitting that change to an endpoint and handling all of that is separate, and when the response comes back, the `data` prop gets updated with whatever new data is present. React will do its magic evaluating what parts of the DOM should get updated when this happens.

packages/cascara/src/ui/Table/TechnicalDesign.mdx
brianespinosa, line 52:
```suggestion
Much thought has been put to the proposal described in this section, our goal is to create a table that can support all of our current use cases while at the same time being abstract enough to support future ones. We are not inventing the wheel here, there are great open source libraries like `react-table`. The problem with those libraries is that they are bloated with things we don't need either because we already abstracted such functionality out to other components or we simply don't need it.
```

brianespinosa: I forgot that Github has this feature to insert a suggested change. WAY easier for some of these changes to just make the edit and suggest it that way. Then you should be presented with a button that allows you to just accept the change.

We will definitely want to use this feature a lot.

packages/cascara/src/ui/Table/TechnicalDesign.mdx
lordpixel: I agree with you, that was an aspect I didn't like too much because of the related `context` thingy there. Just wanted to run this by you to hear your thoughts.

Also, I failed to mention that such extensibility option would be an enhancement. Just in case we haven't covered all actual scenarios. We could also create internal components to fill the gaps as you mention.

I will mmake the corrections to the wording here, thx!

packages/cascara/src/ui/Table/TechnicalDesign.mdx
lordpixel: yes, `atoms` is a more descriptive term.

packages/cascara/src/ui/Table/TechnicalDesign.mdx
lordpixel: yup, that's React Context. I will make it clear throught the document.

packages/cascara/src/ui/Table/TechnicalDesign.mdx
lordpixel: No, this was intended to be dot-spreaded as props in the rendered context.

I was thinking on the `IntentLookup` component that expects the Redux mapped action to be passed, my plan was to dot-spread it on the component passed as `render`.

packages/cascara/src/ui/Table/TechnicalDesign.mdx
lordpixel: ok I will add both apps to the list.

lordpixel:
```suggestion
- Variables
- Status
- Doppio
```

packages/cascara/src/ui/Table/TechnicalDesign.mdx
```diff
@@ -0,0 +1,138 @@
...
+The features in those tables vary from one implementation to another, but the general idea is to present the data in a tabular way, ussually also providing the means for data persistance. The aggregated features list is the following:
+
+**Item creation**:
+Some tables allow the user to create a new entry in the table. Adding an empty row.
+
+**Inline edition**:
+Some tables allow the user to edit the data, replacing the display representation with some JSX that allows the mutation of the data.
+
+**Item deletion**:
+Some tables allow the user to delete record via a button displayed at the row level.
```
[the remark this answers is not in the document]
lordpixel:
```suggestion
Some entities in our API allow item deletion, some others don't. The `delete` functionality must be `opt-in`.
```

packages/cascara/src/ui/Table/TechnicalDesign.mdx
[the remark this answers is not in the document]
lordpixel:
```suggestion
In most implementations, tables add a trailing column to each row. This row can display `action buttons`, which can be mapped to corresponding actions over the entire row. i.e. edit, delete, cancel, etc. 

All of the features mentioned above must be `opt-in`.
```

packages/cascara/src/ui/Table/TechnicalDesign.mdx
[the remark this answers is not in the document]
lordpixel:
```suggestion
The way each cell represents the data for either display or edit purposes not only depends on the data type but also on business logic, in most cases, values are easily coerced to string for display purposes and updates are easily achieved via a controlled input component. The rest of scenarios require a more complex implementation, e.g. the `employee` field in the `Users` table from `Barista Admin` which under the carpet is a single `eid`, but the representation of that cell is an avatar with the user name it points to, although it could be any other representation, specially when editing the value.

Note: The updates made to any row would be notified upstream via `events` so the parent component is able to persist those changes. The means by which the table enables data mutation are to be provided by the developer in the form of `triggers`, the table doesn't provide such means.
```

packages/cascara/src/ui/Table/TechnicalDesign.mdx
```diff
@@ -0,0 +1,138 @@
...
+### Events
+
+In order to support CRUD operations, we must define a mechanism to communicate the parent any changes to the data.
+
+- onCreate
+- onUpdate
+- onDelete
+- onCustomAction
+
+Please note that the names could have been something like `onItemCreate` for semantics, but let's remember that in the future we plan to support selection and custom actions on multiple items. Our proposal is to keep these as abstract as possible.
```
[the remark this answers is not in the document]
lordpixel: @brianespinosa the definitions for the rest of the CRUD events are below.

packages/cascara/src/ui/Table/TechnicalDesign.mdx
```diff
@@ -0,0 +1,138 @@
...
+
+### Config
+
+There will be use cases where features like multi-selection, CRUD, custom actions, etc. won't be required at all, they must be disabled by thefault but easily activated. It makes sense to have boolean props that controll such features.
+
+Another important aspect of configuration is the definition of columns, hence a complete section is devoted to it.
+
+| Name    | Type          | default | Description                                        |
+| ------- | ------------- | ------- | -------------------------------------------------- |
+| Bulk    | Boolean       | `FALSE` | If true, activates the bulk mode (multi-select)    |
```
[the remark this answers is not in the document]
lordpixel: re. `bulk` as an array, it actually specifies if the multipe selection feature is active.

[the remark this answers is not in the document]
lordpixel:
```suggestion
| Bulk (enhancement)    | Boolean       | `FALSE` | If true, activates the bulk mode (multi-select)    |
```

packages/cascara/src/ui/Table/TechnicalDesign.mdx
brianespinosa: @lordpixel Okay, thanks for making this more clear.

1. We want to stay away from anything internal to these components relying on external implementation at all. That means these should not ever need to rely on something like immutable data libraries or state actions from something like Redux. This is because if those external dependencies change or get removed, our components internally could break. These need to work regardless of how that data is getting picked up.
2. Knowing the above... and assuming this ends up being for something like `dataConfig` that is reusable, I wonder if there is a pattern we could use to compose those functions outside of the component where they would be ready to handle this logic as a handler passed into the component. We could have a pretty complex handler function that is supposed to do whatever. The Redux action could be inside that handler function outside of the table. Maybe the onClick function for an action would also return the data relative to where that action was clicked. That way it could be consumed externally in the handler function. I guess in the case of actions, that would mean the onClick would also return the data for the row? 

It's good you put this parameter here. I can now see there are scenarios we would need to know current state of the data at the time of an action firing. And that could only come from inside the component. On this note... I think SUIR is actually doing something similar to this by adding data about the component to React's synthetic events for onClick, etc. Let me go look for some examples of that real quick. It might be the best way to make this more simple. Expose less of the logic internal and still keep a simple external API.

brianespinosa: Yeah, @lordpixel  take a look at this: https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/elements/Button/Button.js#L64 

Maybe we do something similar, but for each of those actions, with onClick on the row, we could return the current data for the row itself, not for the button. This could get tricky when dealing with a row being edited. I guess actually that is okay. It should probably return the actual data for the row, even if it is not being shown based on dataConfig. Just return the object for the row from `data`. If it is a form being edited, the logic for doing that will be contained in the Form itself.

packages/cascara/src/ui/Table/TechnicalDesign.mdx
lordpixel: @brianespinosa, I double-checked the apps to see which atoms are to be included in the MVP.
The lists are the following:

Display
- boolean
- date
- link
- number
- string

Edit
- switch
- input
- datetime picker
- select
- lookup select

The atoms we use for`display` will be different to the atoms we use  for `edit`, that doubles the effort.

packages/cascara/src/ui/Table/TechnicalDesign.mdx
lordpixel: re. #1. and #2.: Not wanting to rely on Redux, just wanted to point out that it was the logic coming from outside, Redux was just an example. This design is completely stack-agnostic.

re. SUIR implementation of events: I've always liked the way an input passes data pretaining to the component itself as well as its current value, we should indeed replicate that pattern, I will keep this in mind when trhowing some code, Thanks!

packages/cascara/src/ui/Table/TechnicalDesign.mdx
brianespinosa: For "edit" the experience needs to be more simple for MVP. I think for boolean we can only do a default html checkbox. For date, we should show an input date time field, not our own date picker. Basically the edit experience for MVP is going to be: whatever is available in html and what do browser vendors provide natively. We will absolutely have other options though in the future:

- switch versus checkbox for booleans
- custom date picker to be consistent across all browsers (except I think keeping native pickers on mobile)
- custom select/dropwdown component of our own so we can control the UI/UX (again I think the exception would be mobile)
- lots of others

packages/cascara/src/ui/Table/TechnicalDesign.mdx
brianespinosa: I see. So not necessarily doing anything with that Redux action itself, but providing it to the action so that we know what the related function is when the action is taken. Another option... is that you could just map on data before passing to the table and add those items to each object in the row if you knew what they were before passing to the data prop. That would work great if we then return the data for that row with each of the action onClick events. You would get back the data for that particular row... which would include the Redux action you passed in to that object when creating the data. If that was the intent... we would need to put a decent example of that implementation pattern together for someone to easily understand that.

```
const newData = originalData.map(data) => data.actions = actionsObjectWithFuncOrWhatever

<Table data={newData} dataConfig={...} />
```

This would essentially eliminate the need for this context prop.

packages/cascara/src/ui/Table/TechnicalDesign.mdx
lordpixel: Ok sounds good! Thanks for reminding me about HTML5, the mobile functionality aspect is very important.

I mentioned the switch instead of checkbox because wee have tables that implement it. At the end of the day we can style an HTML5 checkbox to make it look/behave as a switch.  I have some ideas on that matter that I want to explore and run by @CaffeinatedAllie.

brianespinosa: Yep. I think with these atoms, there are also use cases where we will want a checkbox too. Eventually someone could decide which control they want to show for their boolean. Does a checkbox, a switch, an enum select make the most sense?

packages/cascara/src/ui/Table/TechnicalDesign.mdx
```diff
@@ -0,0 +1,138 @@
...
+| Columns | Array[Object] | `[]`    | Holds the definitions for each column in the table |
+| Create  | Boolean       | `FALSE` | If true, activates the creation of new items       |
+| Delete  | Boolean       | `FALSE` | If true, activates the deletion of items           |
+| Update  | Boolean       | `FALSE` | If true, activates the edit mode                   |
```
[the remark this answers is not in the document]
lordpixel: The description is not very clear on the meaning of `bulk` in this document: instead of `bulk`, I should have used `multiple` to specify if multi-item-selection is enabled because that's what I ment. 

My intention was to have a JSX like this:
`
<Table bulk create update delete />
`

In the above example, `bulk`, `edit`, `create`, and `update` are boolean props that control the `multiple item selection`, `inline creation`, `inline edition`, and `deletion` features. The above example is incomplete though, because if we as developers activate those features, we would also need to consume the events they generate.

A detailed example of what's in my mind is the following:
`
<Table
  // config
  bulk
  create
  delete
  update

  coluns=[{}]

  // events
  onCreate={this.handleNewItem}
  onDelete={this.handleItemDelete}
  onUpdate={this.handleItemUpdate}
  onTrigger={this.handleTriggers}
  
  // triggers
  triggers={[
    { id: 'cancel', callback: this.handleItemCancel, ...additionalConfig },
    { id: 'view', callback: this.handleItemView, ...additionalConfig },
  ]}

  // data
  data={[{}]}
/>
`

The actions you mention are to be passed as `triggers`, the reason I choose that term is for clarity - triggers are pieces of external logic that is `triggered` by our table.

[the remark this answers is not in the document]
brianespinosa: @lordpixel I think this test example is actually something we should also include in the spec. Maybe somewhere towards the bottom. Seems like we are pretty close in terms of honing the props and surface area. We should probably have an area in the doc like this that shows how we think these might work too. This way we can also change it. 

- [ ] Could we add this to the bottom of the doc somewhere and show how some of these might look?

packages/cascara/src/ui/Table/TechnicalDesign.mdx
```diff
@@ -142,3 +142,34 @@ For mutation purposes, the implementation will support `select`, `input`, `boole
...
+
```
brianespinosa, line 147: haha holy crap. That is a lot. This is going to make it difficult to work with our files if we have to jump over big data IDs. I wonder if there is a way to define these big chunks of data as reusable links at the foot of a MDX file. That way if we DO want to do this and not just place a file relative to the file itself, the big chunks of base64 are at the bottom of the document. 

- [ ] Not sure if reusable links will work with base64 images but could we try it?

packages/cascara/src/ui/Table/TechnicalDesign.mdx
lordpixel: As agreed, the pattern we would follow is not the best and presents us with some risks. Removed `render` and `style` props.

packages/cascara/src/ui/Table/TechnicalDesign.mdx
lordpixel: yes, they completely make sense. The enum select will have the `true`, `false` values - we could also allow the user to set the labels to support localizaton.

You just uncovered a track we can start to follow in terms of abstraction, looking forward to it!

lordpixel: I will start typing these ideas into code soon, these kind of convos make really think about implementation! 🙌

packages/cascara/src/ui/Table/TechnicalDesign.mdx
lordpixel: nahh, it didn't work 😞 
but, we now have svg diagrams!

packages/cascara/src/ui/Table/TechnicalDesign.mdx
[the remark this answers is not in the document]
lordpixel: Added!

packages/cascara/src/ui/Table/TechnicalDesign.mdx
```diff
@@ -0,0 +1,170 @@
+---
+name: Spec
+menu: Table
```
CaffeinatedAllie, line 3: Does anyone think that we should do versioning for our specs?

brianespinosa:
Updated with develop. As soon as checks pass again, will merge. Thank you @lordpixel

packages/cascara/src/ui/Table/TechnicalDesign.mdx
```diff
@@ -0,0 +1,138 @@
...
+
+# Proposal
+
+Much thought has been put to the proposal described in this section, our goal is to create a table that can support all of our current use cases while at the same time being abstract enough to support future ones. We are not inventing the wheel here, there are great open source libraries like `react-table`. The problem with those libraries is that they are bloated with things we don't need either because we already abstracted such functionality out to other components or we sinply don't need it.
+
+## Espressive CRUD Table
+
+Given the nature of our actual tables and the operations we need to perform on the data entities, a CRUD Table starts to make sense. A kind of table that not only displays data in a nice tabular way, but also allow for a richier interaction with the data.
+
+It is true that CRUD operations are needed in less scenarios, so the CRUD feature must be easily enabled and must not interfere with the regular table logic.
```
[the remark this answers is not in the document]
lordpixel:
```suggestion
It is true that CRUD operations are needed in less scenarios, so the CRUD feature must be easily enabled and must not interfere with the regular table logic. 
```

docs/src/pages/docs/[[...mdx]].js
```diff
@@ -114,11 +114,16 @@ export const getStaticPaths = async () => {
 };
 
 export const getStaticProps = async ({ params }) => {
-  const mdxDirFiles = getMDXDirFiles(params);
+  const mdxDir = getMDXDirFiles(params);
+
+  // We need to make sure this is only actual files and not a directory
+  // so we are filtering it to make sure the size of the file is not zero.
+  const mdxDirFiles = mdxDir.filter((file) => file.size > 0 && file);
```
Comment (brianespinosa) [inline docs/src/pages/docs/[[...mdx]].js:121]: This is to make sure we are trying to parse MDX for actual MDX files and not a directory.

PR #79, opened 2020-08-28
Feature/table spec

CaffeinatedAllie:
My first pull request!!

My version of the table spec from the design perspective... (also influenced by what Manu did).... Let me know what you think. Obviously I am still trying to work this out.


packages/cascara/src/specs/Table/Table.spec.mdx
```diff
@@ -1,7 +1,155 @@
...
+
 ---
 
 # Table
 
-Let's see if this shows up at our application root.
+## Table Types
+
+- Data table The standard table layout with a table header and rows of data
+- Grid view - more of a gallery view. No table header. Could be considered a list. Need to discuss if we want to make this a separate component or a variation on a table.
```
lordpixel, line 12: Hi @CaffeinatedAllie, from the technical perspective the Grid implementation would be very similar to Table but having it separate might make more sense. Maintaining both together would be hard.

@brianespinosa what do you think?

packages/cascara/src/specs/Table/Table.spec.mdx
```diff
@@ -1,7 +1,155 @@
...
+- Ability to add a data row
+- Ability to edit a table rown (inline optional)
+- Ability to delete a row (optional)
+
+### Table Header
+
+- Column headers with labels and optional descriptions
+- Sortable column headers
+
+### Pagination
```
lordpixel, line 125: This feature already exists as [Pagination](https://github.com/Espressive/cascara/tree/develop/packages/cascara/src/ui/Pagination) component, but it's good to have it here to mark it as `out of scope`.

packages/cascara/src/specs/Table/Table.spec.mdx
```diff
@@ -1,7 +1,155 @@
...
+- Ability to navigate to a specific page of results
+- Ability to navigate to next or previous page of results
+
+### Table Row
+
+- Ability to have cells/columns with specific defined data structures, form fields, action elements (buttons, menus)
+- Ability for results view to be accessible via URL so results can be shared
+
+## Needed/Suggested/Future Table Functionality
+
```
lordpixel, line 139: These are interesting to explore, some of them are already included in the current design, but some others need some more thought. Thanks for adding this section @CaffeinatedAllie, it really opened my mind in terms of architecture. 👍

lordpixel (dismissed):
This is much more detailed than the previous document we had, it really helps understand the variety of Table implementations we have and paves the ways to a really mindful design. 🙌 

Looking forward to the discussion!

packages/cascara/src/specs/Table/Table.spec.mdx
CaffeinatedAllie: Would it be better to describe this as a "list" item.... and when we get to lists we can have multiple formats and grid/gallery view could be one??

packages/cascara/src/specs/Table/Table.spec.mdx
CaffeinatedAllie: Yeah... Maybe we need a related components section. Could we bake in a requirement though to include the pagination component into any implementation of the table component?

packages/cascara/src/specs/Table/Table.spec.mdx
CaffeinatedAllie: YAYYY!! I think talking big picture/future enhancements right now will really help future proof things. I am glad that what I am doing is helping you!

packages/cascara/src/specs/Table/Table.spec.mdx
brianespinosa: @lordpixel hard to say yet. Initially, yes I think it should be a separate component for managing that different display experience. The public API pertaining to the data input and configuration might be the same, but other display props might be different that control the experience. If the input props are 1:1 then we could have the overall admin layout experience later on where you could switch between display views if you want. But the implementation of that might be left to the developer initially.

@CaffeinatedAllie I think it is good we are thinking about that.

IMO, let's define that as a separate component for now and we can evaluate working it into an overall layout where switching views "just works" later.

packages/cascara/src/specs/Table/Table.spec.mdx
brianespinosa: - [ ] Yes. Let's add a "Related Components" section for now while we figure this out. This will allow us to define separate area needs in the spec without having to flesh them out.

packages/cascara/src/specs/Table/Table.spec.mdx
brianespinosa: @lordpixel this brings up a good point for us, glad we are calling it out here. Whatever we implement in terms of query params being the state of the filter and pagination... we should make sure that tooling is independent of whatever routing tools we create.

This sounds like a great example of a custom hook. Maybe something like `useQueryState` which composes `useEffect`. This could basically be a value and function. Value would be an object of the current query params as key/values. The function would trigger setting those values based on the object returned. That would make it so we could manage query params 100% independent of router. Then it would work just like useState but it would essentially hoist that state into the url as query params.

There are a number of other areas we will need this logic too.

brianespinosa: Funny... did a quick search and found this. I do not like that it requires a context provider. I think this could be done as a side effect and be more agnostic to router... but looking into the code, I can see why this does make sense NOT being a side effect and doing this in a more React way. We should bookmark this and come back to it. https://github.com/pbeshai/use-query-params

packages/cascara/src/specs/Table/Table.spec.mdx
```diff
@@ -1,7 +1,155 @@
...
+- Table header - horizontal VS vertical
+- Ability to customize columns visible
+- Ability to reorganize position/order of columns
+- Ability to have header/columns be sticky
+- Accordion columns
+- Nested tables
+- Tables with a tree like structure
+- Way to handle truncated/text that is too long to fit
+- Mobile view
+- Ability to set a default sort on a column
```
brianespinosa, line 151: @CaffeinatedAllie @lordpixel IMO this should actually not be a feature of the table itself. We pass data into the table and then tell the table what to display. So if we want to change that, I think the best way to handle that is to change the array of `data` being passed into the table.  That is going to lead to less confusion I think in terms of how that data displays. If we are changing the order inside of the table itself, that becomes more complex in terms of figuring it out. We can have some utils for handling that external from the table.

@lordpixel that would also mean that we will need to define a way for hooking this functionality up to table headers. When we define dataConfig we may want to specify a `sorting` attribute for each object. Enums could be `asc` or `desc`. These would be for display only so the header could indicate which cell was doing the sorting and in what direction. Each header cell would have a click handler that could fire whatever func was passed in as part of that dataConfig object as well. Essentially the table will then just display what it is told in the header row too. And when clicked, fire whatever associated function is needed. Sorting of data can happen external. And if there is any strange Array.map work we need to do to results externally can also happen outside of Table.

brianespinosa: -[ ] @lordpixel @CaffeinatedAllie I think we remove this line? Or if we leave it, make a note that the header cells should be clickable and DISPLAY a sort order for a column?

packages/cascara/src/specs/Table/Table.spec.mdx
```diff
@@ -1,7 +1,155 @@
...
+- Ability to have tooltips on table header columns/table cells
```
brianespinosa, line 152:
```suggestion
- [P3?] Ability to have tooltips on table header columns/table cells
```

packages/cascara/src/specs/Table/Table.spec.mdx
```diff
@@ -1,7 +1,155 @@
...
+- Hover affect for table rows/columns
```
brianespinosa, line 153:
```suggestion
- [MVP] Hover affect for table rows/columns
```

packages/cascara/src/specs/Table/Table.spec.mdx
```diff
@@ -1,7 +1,155 @@
...
+
+- Ability to retain last/multiple/persistent configuration
```
brianespinosa, line 155: We DEFINITELY want things like this with Cascara. We need to evaluate how we want this to happen though so it is consistent in Cascara. Some of these things might make sense to persist to the user data. Some might make sense to persist to local storage if there is incomplete work being done like on a form.

```suggestion
- [P3] Ability to retain last/multiple/persistent configuration
```

packages/cascara/src/specs/Table/Table.spec.mdx
lordpixel: They read your mind @brianespinosa! ok I'll take a look, thanks!

packages/cascara/src/specs/Table/Table.spec.mdx
```diff
@@ -1,7 +1,155 @@
...
+
+### Table Row
+
+- Ability to have cells/columns with specific defined data structures, form fields, action elements (buttons, menus)
+- Ability for results view to be accessible via URL so results can be shared
+
+## Needed/Suggested/Future Table Functionality
+
+- Zebra striping on rows
+- Ability to bulk edit, delete, add records
```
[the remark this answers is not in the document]
lordpixel: This is great, I'll add that to the specs as well.

packages/cascara/src/specs/Table/Table.spec.mdx
```diff
@@ -1,7 +1,155 @@
...
+- Table header - horizontal VS vertical
+- Ability to customize columns visible
```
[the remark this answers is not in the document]
lordpixel: This is doable with the current spec.

packages/cascara/src/specs/Table/Table.spec.mdx
```diff
@@ -1,7 +1,155 @@
...
+### Table Row
+
+- Ability to have cells/columns with specific defined data structures, form fields, action elements (buttons, menus)
+- Ability for results view to be accessible via URL so results can be shared
+
+## Needed/Suggested/Future Table Functionality
+
+- Zebra striping on rows
+- Ability to bulk edit, delete, add records
+- Table header - horizontal VS vertical
```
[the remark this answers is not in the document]
lordpixel: I did some experiments with nested, multi-level, vertical headers. Vertical headers work best for tables with a fixed number of rows, looking forward for this feature.

packages/cascara/src/specs/Table/Table.spec.mdx
```diff
@@ -1,7 +1,155 @@
...
+- Ability to customize columns visible
+- Ability to reorganize position/order of columns
```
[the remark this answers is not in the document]
lordpixel: This would require support for drag and drop, I would expect this to be in P2. Can we move it?

packages/cascara/src/specs/Table/Table.spec.mdx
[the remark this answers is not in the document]
lordpixel: The current spec was written with these features in mind, I am positive that we can have these in the MVP.

packages/cascara/src/specs/Table/Table.spec.mdx
```diff
@@ -1,7 +1,155 @@
...
+- Ability to navigate to next or previous page of results
+
+### Table Row
+
+- Ability to have cells/columns with specific defined data structures, form fields, action elements (buttons, menus)
+- Ability for results view to be accessible via URL so results can be shared
+
+## Needed/Suggested/Future Table Functionality
+
+- Zebra striping on rows
```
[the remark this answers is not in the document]
lordpixel: This is really simple CSS, we can have it in the MVP. What do you think?

packages/cascara/src/specs/Table/Table.spec.mdx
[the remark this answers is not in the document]
lordpixel: I did some experiments with vertical, nested headers. These work best with tables that have a fixed number of rows, looking forward to see this feature spected.

packages/cascara/src/specs/Table/Table.spec.mdx
```diff
@@ -1,7 +1,155 @@
...
+- Ability to bulk edit, delete, add records
+- Table header - horizontal VS vertical
+- Ability to customize columns visible
+- Ability to reorganize position/order of columns
+- Ability to have header/columns be sticky
+- Accordion columns
```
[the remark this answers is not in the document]
lordpixel: This could be P2, if we move what we can to MVP.

packages/cascara/src/specs/Table/Table.spec.mdx
```diff
@@ -1,7 +1,155 @@
...
+- Nested tables
+- Tables with a tree like structure
+- Way to handle truncated/text that is too long to fit
+- Mobile view
```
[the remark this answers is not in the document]
lordpixel: ok

packages/cascara/src/specs/Table/Table.spec.mdx
lordpixel: @brianespinosa we can do the following:

1. add the sort definition in the column schema and supporting enums
2. expose an event `onSort` or something similar with the sig. (attribute, direction) => {}

If the no sort config is passed to the column, the buttons in the header won't render.

packages/cascara/src/specs/Table/Table.spec.mdx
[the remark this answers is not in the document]
brianespinosa: Copy paste error on my part. I did intend this for MVP. Will update my suggestion. Thanks @lordpixel

packages/cascara/src/specs/Table/Table.spec.mdx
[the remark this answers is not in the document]
brianespinosa: @lordpixel I want to make sure we draw the distinction between something that is kind of working, versus something fully working. I think we should not consider editing rows fully released as MVP even though it will work. When we do P2, we want to invest some time in adding all of the other Atom types we want for different fields display/editing. And also make sure the editing experience is optimal across all fields. Things like trimming values onBlur, onSubmit, etc.

But you do bring up a good point here too. We may want to communicate how some of these features work early on. And maybe if we do put a release out, if there is functionality that is still experimental, maybe name all vars related with `unsafe_` until they are fully supported. This way we can have something released with the rest of Cascara, but have some new things that are risky to touch.

packages/cascara/src/specs/Table/Table.spec.mdx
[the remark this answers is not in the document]
brianespinosa: Good that we are thinking about this. Using Grid this should theoretically be pretty easy to achieve. My concern on this one is that we actually do not have this type of display currently. We should evaluate what scenarios we would need this. @CaffeinatedAllie we might determine we do not need this pattern anyway.

@CaffeinatedAllie this is also making me think that these documents may end up needing to be something we review and assign specific needs for each one of these. There are a couple I think we may not need.

packages/cascara/src/specs/Table/Table.spec.mdx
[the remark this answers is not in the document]
brianespinosa: This is another scenario where I suspect this might be a pattern we actually do not need. Not yet seeing the use cases where it would be needed. But good to have these items here anyway so we can evaluate if our public API would allow us to achieve it if needed later on.

packages/cascara/src/specs/Table/Table.spec.mdx
brianespinosa: @lordpixel I think we could make it even more simple. Having discreet onEVENT functions decorating the Table forces us down specific paths. If we instead give more implementation agnostic patterns to developers, they can potentially do more.

`dataConfig` could have these params included in each object:

```
{
  sorting: 'asc',
  onHeaderClick: handleDataSortAndStuff()
  ...
}
```

The onHeaderClick would be mapped to an onClick function on the header cells. That handler function could do all kinds of things outside of the Table. It could set and remove `sorting` attributes to other objects in `dataConfig` and then change them when the user clicks. It could also change data in `data` being passed to the table. These changes will cause the table to render and update. This can be achieved without any `onSort` function on the Table itself.

From a DX perspective... maybe there will appear some consistent scenarios where we want to expose some `onSOMETHING` functions on these components. I think my worry is that it might still be early to do that in a consistent way. Whatever functions we expose on the tables, I think we want to make sure they are consistent from one component to the next. And if we force ourselves to follow some consistent React composition patterns first, I think we will have fewer bugs and impossible states to fight with.

PR #80, opened 2020-08-28
Added button spec

CaffeinatedAllie:
Rough draft of button spec... can flesh out more once I get suggestions from the team :)


packages/cascara/src/specs/Button/Button.Spec.mdx
```diff
@@ -0,0 +1,49 @@
...
+
+---
+
+# Button
+
+A button is an atom level element that is used alone as well in a variety of larger components.
+
+## Existing Button Types
+
+- Priority/Intention (not clearly defined, but implied)
```
lordpixel, line 13: We'll use these types in table for the CRUD buttons.

packages/cascara/src/specs/Button/Button.Spec.mdx
```diff
@@ -0,0 +1,49 @@
...
+	- Primary
+	- Secondary
+	- Tertiary
+	- Text only
+	- Destructive
+- Icon
+	- With icon
+	- Icon only
+- Boolean button
```
lordpixel, line 22: It seems that the Boolean button could be converted into a checkbox that is displayed as a switch, visually it is more descriptive. What do you think @CaffeinatedAllie and @brianespinosa?

CaffeinatedAllie: We already have buttons that also act as a toggle.... think the UI element where the user picks the day of the week. An on/off switch would be awkward for that kind of UI element.

PR #81, opened 2020-08-28
Feature/checkbox spec

CaffeinatedAllie:
Added spec for checkboxes


packages/cascara/src/specs/Checkbox/Checkbox.Spec.mdx
```diff
@@ -0,0 +1,39 @@
+---
+title: Button
```
lordpixel, line 2: This should be `Checkbox`

CaffeinatedAllie: DOH!

PR #85, opened 2020-09-01
Docs prop table support

brianespinosa:
Very lightweight. This at least adds the ability to define a prop table if we need it. Not the best presentation yet but we can come back to it.

Also adds conditional structure support on Admin for a drawer component. Not responsive yet.


brianespinosa:
image

PR #89, opened 2020-09-02
Set up a Public API spec example fixture proposal

brianespinosa:
This is an exploration on how we might want to explore a public API before we dig further into implementation details. This might be something good to do in addition to a technical writeup. Might even make sense to do the whole technical part this way if we can make some more helper tooling to throw these prototypes together quickly like this. Though, with Cosmos and the new JsonDisplay component... this was pretty fast.


lordpixel (approved):
this looks good, so not so many props?
I was in the understanding that we'll pass some event handlers.

brianespinosa:
> this looks good, so not so many props?
> I was in the understanding that we'll pass some event handlers.

This is the difference I think I have been failing to illustrate. Which made me realize I think we need to assemble some working examples of things this way to evaluate how we might work with this data inside of a component. The "prepared data" is essentially what we would render into the row.

There might be other event handlers, and there might not. We can take a look at this together and evaluate what sorts of use cases might be problematic with this. If we reach something that is too restrictive, we can then extend the API further to accommodate any problematic use cases.

I think we should also aim to push this direction in terms of this public API pattern having more semantically global feeling patterns, since this same sort of pattern can be applied to other components in Cascara. This is another larger picture I have failed to communicate, but I think working through this example may help with that... because I think from a config/data perspective, these examples should also be thought experiments for lists, forms, etc. I can walk us through this together in our zoom.

PR #90, opened 2020-09-04
table markup

lordpixel:
This adds the a Table fixture to explore markup


packages/cascara/src/ui/Table/TableStyleTest.fixure.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import faker from 'faker';
-import './TableStyleTest.module.scss';
+//import './TableStyleTest.module.scss';
```
brianespinosa, line 3: @lordpixel was this intentional disabling the style import in the table style fixture?

packages/cascara/src/ui/Table/Table.fixture.js
```diff
@@ -0,0 +1,104 @@
...
+              ])}
+          </tr>
+        ))}
+      </tbody>
+      <tfoot>{<tr>{columns}</tr>}</tfoot>
+    </table>
+  );
+};
+
+const Fixture = <Table config={dataConfig} data={fakeEmployees} />;
```
brianespinosa, line 102: Sweet!

packages/cascara/src/ui/Table/mockData/index.js
```diff
@@ -0,0 +1,23 @@
...
+      eid: faker.random.uuid(),
+      fullName: faker.fake(
+        '{{name.lastName}}, {{name.firstName}} {{name.suffix}}'
+      ),
+      homePhone: faker.phone.phoneNumber(),
+      nickname: faker.name.firstName(),
+      officePhone: faker.phone.phoneNumber(),
+      title: faker.name.jobTitle(),
+    });
+  });
```
brianespinosa, line 23: I love that you went ahead and did this BTW. We should talk in a month and see what other needs we might have for generating consistently seeded, random data. There are going to be other needs we have too. Might be good to invest a little more in some good tooling for faster prototyping.

packages/cascara/src/ui/Table/style.css
```diff
@@ -0,0 +1,22 @@
...
+  position: relative;
+  border-collapse: collapse;
+}
+th,
+td {
+  padding: 0.25rem;
+}
+th {
+  background: white;
+  position: sticky;
```
brianespinosa, line 12: I saw in your example this is working btw.

When. I tried doing this a little over a year ago, when doing position on any part of the table, it removed it from the flow of the document and the column widths were no longer tied to the width of the columns below in the table. It I think was not a problem in all browsers. I do not recall which worked or did not. Might have only been wrong in Chrome. Seems good now though.

brianespinosa (approved):

PR #91, opened 2020-09-07
Table markup fixture

lordpixel:
# Table Context
A table context has been added, it currently handles selection and it's the source of data for all rows.
A new row context will be created to handle row logic.

## Semantic chunks
This fixture separates the table into more semantic chunks: Table, ActionBar, Header, Rows and Cells. Our goal is to have a discussion on whether this structure and the chunks it proposes are what we want.

## Header and ActionBar stickyness
The design for Table in InVision shows the `ActionBar` below the `Header`, since the header must be sticky, the following questions arise:

- do we want the ActionBar to be sticky as well?
- if `ActionBar` must be sticky, there is a challenge on how to prevent it from overlaping the `Header`

In regards to the last point above, please see the cosmos fixture and scroll the table. Notice how we had to create a new stacking context so `Header` and `ActionBar` can pass each other. I consider this as a hack that opens the door to nasty bugs.

## CSS Grid and Column (re)sizing
Since we are using table markup, we can leverage `<col/>` tag as the means of resizing the columns. The column size will then be translated to `CSS Grid`'s `grid-template-columns` property. What do you think? is this worth to spend some time exploring? if not, what would be the approach to follow here?

## Selection
The global selection toggle, needs to display the `indeterminate` property in partial selection scenarios. Any guidance on this would be greatly appreaciated.


brianespinosa (approved):
Awesome!

PR #92, opened 2020-09-08
Docs now support index and principles section

brianespinosa:
## New Component Checklist


lordpixel (dismissed):
🙌

PR #102, opened 2020-09-11
Table - split HTML markup into React components

lordpixel:
This PR separates the previous fixture into React components. It also adds an `Export` fixture for demonstrating `bulk actions`.


packages/cascara/src/ui/Table/Table.js
```diff
@@ -0,0 +1,35 @@
...
+
+import TableContext from './context/TableContext';
+
+import TableHeader from './TableHeader';
+import TableBody from './TableBody';
+
+const Table = () => {
+  const {
+    dataConfig: { actions, bulkActions, display },
+  } = useContext(TableContext);
```
brianespinosa, line 11: It looks like right now all the data is getting defined directly in the table context? How are end users getting data into that context? I don't think we want to export contexts from Cascara. We should probably be exposing props on this Table and then passing that into context within this component. Unless this Table is not actually the official component where that happens?

packages/cascara/src/ui/Table/TableHeader.js
```diff
@@ -0,0 +1,83 @@
+import React, { useContext } from 'react';
+
+import Button from '../Button';
+import DownloadButton from '../../modules/DownloadButton';
+
+import TableContext from './context';
+import SelectionToggle from './atoms/SelectionToggle';
+import ActionBar from './ActionBar';
+
+const TableHeader = ({ footer }) => {
```
brianespinosa, line 10: I think we should actually make TableHeader and TableFooter separate components entirely. The reason for this is eventually the styles are going to be tied to these two areas directly with modules. And the logic inside them may be different. 

- [ ] Maybe set up a separate TableFooter component?

packages/cascara/src/ui/Table/fixtures/TableStyleTest.module.scss
```diff
@@ -1,10 +1,11 @@
 :global {
```
brianespinosa, line 1: This is not how the styles are actually going to be attached to all of these table elements. This was initially just for that test. We can still leave this way for now... but we actually want to update these to apply them directly as CSS Modules to each of the table elements.

These items should probably also get renamed with classes to set those styles accordingly. That can be a separate task though: https://espressive.atlassian.net/browse/DEV-12472

packages/cascara/src/ui/Table/index.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+
+import { TableContextProvider } from './context/TableContext';
+import Table from './Table';
+
+export default ({ data, dataConfig, onAction }) => (
+  <TableContextProvider data={data} dataConfig={dataConfig} onAction={onAction}>
+    <Table />
+  </TableContextProvider>
+);
```
brianespinosa, line 10: I see... we are using the index.js to set the parameters.

This might present some issues for us in terms of how the exporting and naming of the components work. For instance, our doc typing I do not think will work with this if we define `index.js` as the file to run typing on. The docgen I think will also fail because I do not think exporting default from an index file will apply a component name. 

Maybe this will be okay if we define a component name? I think the problem with that is we are defining the index as Table and then we also have a component called Table. This is likely to be a confusing pattern for devs. 

- [ ] Let's talk about this together and figure out what the pros and cons of this pattern are.

brianespinosa (changes requested):

brianespinosa:
@lordpixel I made a task on all items that I _think_ need to be resolved on this PR. Anything that can be done separately, I went ahead and made a Jira task for it so it does not block this PR.

.vscode/settings.json
```diff
@@ -25,20 +25,6 @@
...
-    "activityBar.inactiveForeground": "#e7e7e799",
-    "activityBarBadge.background": "#48b010",
-    "activityBarBadge.foreground": "#e7e7e7",
-    "statusBar.background": "#850c37",
-    "statusBar.foreground": "#e7e7e7",
-    "statusBarItem.hoverBackground": "#b4104a",
-    "titleBar.activeBackground": "#850c37",
-    "titleBar.activeForeground": "#e7e7e7",
-    "titleBar.inactiveBackground": "#90023A99",
-    "titleBar.inactiveForeground": "#e7e7e799",
```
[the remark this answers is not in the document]
lordpixel: This change was generate from `esp-app-web` somehow, let me put it back.

packages/cascara/package.json
```diff
@@ -23,6 +23,7 @@
   },
   "devDependencies": {
     "@espressive/legacy-css": "file:../../utils/legacy-css",
+    "moment": "2.27.0",
```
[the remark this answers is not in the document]
lordpixel: this is a dumb mistake on my part, it was not my intention to add it to this project but somehow I did. Thanks for the catch!

packages/cascara/src/index.js
```diff
@@ -1,6 +1,7 @@
 import Admin from './structures/Admin';
 import Button from './ui/Button';
+import DownloadButton from './modules/DownloadButton';
```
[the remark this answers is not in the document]
lordpixel: agreed, will move this to modules once it's good to go.

packages/cascara/src/ui/Table/ActionBar.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+
+const ActionBar = ({ actions = [], title = '' }) => (
+  <>
+    {title && <span>{title}</span>}
+    {actions.length && actions}
+  </>
+);
+
+export default ActionBar;
```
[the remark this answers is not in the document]
lordpixel: Yup, it is used in both scopes.

packages/cascara/src/ui/Table/index.js
lordpixel: yup this is not a good pattern, I didn't want to spend time on fixing it because it was just a fixture.
Let me finish the convertion to React components and fix this, thanks for the catch!

packages/cascara/src/ui/Table/Table.js
lordpixel: You are right, Table is wrapped by the TableContext. The index in `packages/cascara/src/ui/Table` exports the

packages/cascara/src/ui/Table/TableHeader.js
lordpixel: sure thing, we will have those as separate components.

.vscode/settings.json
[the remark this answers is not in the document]
brianespinosa: @lordpixel you put these back into the extensions.json file. The extensions.json file controls what extensions VSCode will suggest to people when opening the editor.

- [ ] Add settings back here

.vscode/extensions.json
```diff
@@ -1,12 +1,31 @@
...
-    "andrewleedham.vscode-css-modules",
-    "bungcip.better-toml",
-    "clinyong.vscode-css-modules",
-    "dbaeumer.vscode-eslint",
-    "mgmcdermott.vscode-language-babel",
-    "silvenon.mdx",
-    "xyc.vscode-mdx-preview"
-  ],
-  "unwantedRecommendations": []
-}
```
brianespinosa, line 12: - [ ]  Please restore the extensions.json file

brianespinosa (changes requested):

.vscode/extensions.json
lordpixel: This is caused by vscode.

packages/cascara/src/ui/Table/index.js
lordpixel: The change has been applied, Table is the context container and the index only exports table.

brianespinosa (approved):

brianespinosa:
@lordpixel there are a few conflicts with develop. As soon as those get resolved, this is officially approved to merge.

It looks like there was a change to that test CSS file that was adding body padding to ALL fixtures (another reason we do not want to add global type styles. ;)

brianespinosa (approved):

PR #109, opened 2020-10-12
[FDS-13, FDS-23, FDS-24, FDS-25] - Table P1

lordpixel:
Hi team,

This PR is still in progress, expect more commits. The goal of this draft is to share the daily progress and open a channel to discuss the ongoing implementation.

Looking forward to hear your comments!


packages/cascara/src/lib/mock/generateFakeEmployees.js
```diff
@@ -6,15 +6,18 @@ export const generateFakeEmployees = (qty) =>
   new Array(qty).fill(null).map((e, i) => {
     // So we never have collisions, we have to start at 2 instead of 1
     faker.seed(i++ + 1);
+
+    const firstName = faker.name.firstName();
+    const lastName = faker.name.lastName();
+
     return (e = {
```
lordpixel, line 13: What we want with this fixture data is to have one field of each supported type.

packages/cascara/src/ui/Table/TableFooter.js
```diff
@@ -1,84 +1,53 @@
...
+  const { dataConfig } = useContext(ModuleContext);
+
+  const actionBarCell = (
+    <ActionBar
+      actions={dataConfig.actions.map((action) => {
+        let Module;
+
+        switch (action.module) {
+          case 'edit':
+            Module = ActionEdit;
```
lordpixel, line 20: This is the case where we need to define a better way for mapping `dataType` -> `Module`.

packages/cascara/src/ui/Table/TableRow.js
```diff
@@ -1,59 +1,106 @@
...
       <ActionBar
-        actions={actions.map((action) => (
-          <Button
-            {...action}
-            content={action.label}
-            key={action.label}
-            onClick={() => handleOnAction(action, id)}
-          />
-        ))}
+        actions={dataConfig.actions.map((action) => {
```
lordpixel, line 29: Here, the need of a `DataType` -> `Module` mapping rises again.

brianespinosa (approved):
I think these are good to go. In the future, we do not want to have PRs this large I don't think. That is probably more my fault in terms of separating these things out. Also, I think making sure that stories can be worked on separately and then get approved along the way will make things easier.

For now, I think this is just the nature of where we are. As we get more components and work in Cascara, it will be easier to work on things in smaller chunks.

packages/cascara/src/ui/Table/TableRow.js
```diff
@@ -54,7 +54,6 @@ const TableRow = ({ config = {}, record = {} }) => {
     const { module: type, ...rest } = column;
     let Module;
 
-    debugger;
     switch (type) {
```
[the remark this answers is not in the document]
lordpixel: oh yes, the original name is `module`, it is renamed while destructuring. I did it for readability but you are right, let me change it.

packages/cascara/src/modules/ActionButton/ActionButton.js
```diff
@@ -2,16 +2,30 @@ import React, { useContext } from 'react';
 import { ModuleContext } from '../context';
 import { Button } from 'semantic-ui-react';
 
-const ActionButton = ({ content = 'ActionButton', ...rest }) => {
+const ActionButton = ({
+  content = 'ActionButton',
+  isLabeled = false,
+  label,
```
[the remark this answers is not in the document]
lordpixel: yeah, I didn't like the `content` prop for the same reasons, `label` is mmore semantic.

PR #120, opened 2020-12-02
DEV-13327 - fix Table action name is not reported correctly by onAction

lordpixel:
Hi team,

The action name for the`edit` module was being reported as `actionName`, which is incorrect, the correct term is `name`. Also, the documentation was improved to be more clear in terms of how to use Actions.

cheers


brianespinosa (approved):
@lordpixel this LGTM. I do not think my other PR negatively impacts this anywhere either.

PR #121, opened 2020-12-02
FDS-54 Scrollbar style fixes, FDS-52 Sticky actions

brianespinosa:
FDS-54
Cleans up some of prototype styles for scrolling in tables and makes scroll areas auto so they do not show up all the time, and when they do, they now use the defined scroll styles for Cascara. When the styles for scroll get updated in Cascara, these scroll bar styles will also now get updated.

FDS-52
Makes action buttons sticky inside of a table so they will always be visible inside a table even in a long horizontal view. This also cleans up the buttons that were defined for Action Modules. Previously Button and Edit were defined with SUIR Buttons. These were switched to Reakit buttons using the SUI classnames for correct styles. Additionally, this solves a prop and classname conflict that was happening for the Button component that had a 'label' prop attached to it. That label prop changes the overall styling of the button because the concept of "labels" in SUI is a visual treatment of attaching a rounded box to an input or button. Reakit buttons do not throw strict mode warnings on us either.

@lordpixel please check my work here on the Edit and Button action modules regarding the above updates.


lordpixel (dismissed):
sweet 👍

lordpixel (approved):

brianespinosa:
The Firefox issue is becoming a little tricky. Going to work on that in a separate PR.

PR #124, opened 2020-12-11
FDS-63 Update legacy CSS to resolve some long-standing issues

brianespinosa:
This should fix the following:

- Inter font now part of legacy CSS dependencies and @font-face
- Input labels no longer with a left margin indentation (should be flush left aligned with the input box)
- Checkboxes are no longer :swear: round
- Radio boxes are no longer smaller than checkboxes

We will ship a new `@espressive/legacy-css` package that can be used in all apps once this is approved and merged.

image


lordpixel (approved):
:beers:

utils/legacy-css/src/css/semantic.min.css
```diff
@@ -1,3 +1,6 @@
+/* This tilde syntax is used by Webpack to resolve the "inter-ui" node package in node_modules. */
+@import '~/inter-ui/inter.css';
```
lordpixel, line 2: we may need to add the alias to webpack config, which comes from CRA.

brianespinosa: Swapped this out. We don't want to use a Webpack syntax if we can avoid it. I was kind of hoping this would be okay until we got rid of them. But it dies in Next.

The updated solution references the direct node_modules path of the CSS which should work in all environments, regardless of bundler. 👏🏽

PR #127, opened 2020-12-22
FDS-91 - Fix checkbox not resetting to the original value if canceled

lordpixel:
The checkbox module in `display` mode was using the `isChecked` value from its `local state`, which is incorrect. The value to  `display` should always come from the actual data,  as it is the only source of truth.

x8Gr47Uy9A

This PR fixes the bug by implementing the `useCheckboxState` hook for the checkbox' state as well as other minor changes.

FDS-91 fix

During our pair-programing session we discovered some other issues which we'll address in future PRs, once the relevant tasks are created.

### DIFF
--- a/packages/cascara/src/modules/DataCheckbox/DataCheckbox.js
+++ b/packages/cascara/src/modules/DataCheckbox/DataCheckbox.js
@@ -1,8 +1,8 @@
 import React, { useContext } from 'react';
-import { Checkbox } from 'reakit/Checkbox';
+import { Checkbox, useCheckboxState } from 'reakit/Checkbox';
 import pt from 'prop-types';
+
 import { ModuleContext } from '../context';
-import useToggle from '../../hooks/useToggle';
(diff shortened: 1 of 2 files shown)


packages/cascara/src/modules/DataCheckbox/DataCheckbox.js
```diff
@@ -36,20 +36,18 @@ const DataCheckbox = ({
     attribute && record
       ? getAttributeValueFromRecord(attribute, record)
       : value;
-  const [isChecked, setIsChecked] = useToggle(Boolean(finalValue));
+  const checkbox = useCheckboxState({ state: Boolean(finalValue) });
```
brianespinosa, line 39: @lordpixel I think it is best that we leave out the other work of making these modules more simple and getting the value from the record and passing it in at the Body/Row level in a separate PR. Thank you for leaving that out here so we can get the fix in first. 👍🏽

brianespinosa (approved):

PR #131, opened 2021-01-06
FDS-75 - use node v12.20.0 (LTS)

lordpixel:
Hi team,

Nodejs has been set to v12 for the monorepo as well as any child project/config that also enforces a specific version.

Smoke tests were performed to all commands for the monorepo such as:
`
$yarn install
`
`
$yarn cascara build
`
`
$yarn docs start
`
`
$yarn docs build
`
`
$yarn cosmos
`
`
$yarn cosmos:build
`
`
$yarn test
`
`
$yarn lint
`
`
$yarn tokens
`

Could you please double check everything works OK in case I've had missed something?

Thanks


rollup.config.js
```diff
@@ -84,7 +84,7 @@ export const getRollupConfig = ({ pwd, babelConfigFile }) => {
   const esConfig = {
     input,
     output: {
-      dir: `${SOURCE_DIR}/${pkgConfig?.module?.replace('/index.js', '')}`,
+      dir: `${SOURCE_DIR}/${pkgConfig.module.replace('/index.js', '')}`,
```
brianespinosa, line 87: @lordpixel was there a reason we needed to remove the optional chaining in Node 12 here? I think this syntax was used for packages that might not necessarily have a `module` parameter declared.

brianespinosa (changes requested):
One change and a question. But I do not think these should be major. This should be ready to merge pretty quick.

docs/package.json
```diff
@@ -9,7 +9,7 @@
   },
   "dependencies": {
     "@espressive/cascara": "file:../packages/cascara",
-    "@espressive/legacy-css": "latest",
+    "@espressive/legacy-css": "2.0.1",
```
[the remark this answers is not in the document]
lordpixel: This kills the build, that's why I used the specific version.

rollup.config.js
lordpixel: The elvis op. sees not to be supported. Rollup didn't like it.

PR #137, opened 2021-01-13
FDS-95 - Prevent multiple rows from entering in edit mode

lordpixel:
Hi team,

As of today, all reccords in a Table can be edited at the same time, which causes lots of issues. The root cause seems to be that the `isEditing` flag is scoped to the row, so every row is in charge of its edit state.

**The bug**
https://recordit.co/U8HRrG52bJ

The flag was replaced with the actual id of the record that is being edited, we also moved the flag and its setter functions to the upper `context`.

**The fix**
https://recordit.co/J6b3d6NLCj

I am sure there could be other ways to implement this, please feel free to propose any other approach.

Cheers.

### DIFF
--- a/packages/cascara/src/modules/ActionEdit/ActionEdit.js
+++ b/packages/cascara/src/modules/ActionEdit/ActionEdit.js
@@ -17,11 +17,20 @@ const propTypes = {
 };
 
 const ActionEdit = ({ dataTestIDs, editLabel = 'Edit' }) => {
-  const { isEditing, setIsEditing, formMethods, record, onAction } = useContext(
-    ModuleContext
-  );
+  const {
+    idOfRecordInEditMode,
+    isEditing,
+    enterEditMode,
+    exitEditMode,
+    formMethods,
+    record,
+    onAction,
+    uniqueIdAttribute,
+  } = useContext(ModuleContext);
   const { handleSubmit, formState, reset } = formMethods;
   const { isDirty, isSubmitting } = formState;
+  const recordId = record[uniqueIdAttribute];
+  const whenAnotherRowIsEditing = Boolean(idOfRecordInEditMode);
 
   /**
    * this seems like ugly, we need to find a better way
@@ -47,7 +56,7 @@ const ActionEdit = ({ dataTestIDs, editLabel = 'Edit' }) => {
       }
     );
 
-    setIsEditing(false);
+    exitEditMode();
   };
 
   const handleCancel = () => {
@@ -72,7 +81,7 @@ const ActionEdit = ({ dataTestIDs, editLabel = 'Edit' }) => {
       }
     );
 
-    setIsEditing(true);
+    enterEditMode(recordId);
   };
 
   const onSubmit = (data) => {
@@ -87,7 +96,7 @@ const ActionEdit = ({ dataTestIDs, editLabel = 'Edit' }) => {
       }
     );
 
-    setIsEditing(false);
+    exitEditMode();
   };
 
   return isEditing ? (
@@ -115,6 +124,7 @@ const ActionEdit = ({ dataTestIDs, editLabel = 'Edit' }) => {
     <Button
       {...editTestId}
       className='ui basic button'
+      disabled={whenAnotherRowIsEditing}
       onClick={handleEdit}
       type='button'
     >
--- a/packages/cascara/src/ui/Table/context/RowProvider.js
+++ b/packages/cascara/src/ui/Table/context/RowProvider.js
@@ -4,16 +4,24 @@ import { ModuleContext, ModuleProvider } from '../../../modules/context';
 const RowProvider = ({ children, value, ...props }) => {
   const grandparentValues = useContext(ModuleContext);
 
-  // Here we need to not pass the isEditing and setIsEditing values from grandparent which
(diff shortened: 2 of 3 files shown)


packages/cascara/src/modules/DataCheckbox/DataCheckbox.js
```diff
@@ -30,8 +30,14 @@ const DataCheckbox = ({
   value,
   ...rest
 }) => {
-  const { isEditing, formMethods } = useContext(ModuleContext);
+  const {
+    formMethods,
+    record,
+    uniqueIdAttribute,
+    idOfRecordInEditMode,
+  } = useContext(ModuleContext);
```
brianespinosa, line 38: @lordpixel I think for this solution, we want the Data Modules to not have to be aware of all this information. Remember... they will also be used inside of a Form.

They should be able to look up the the nearest Module Provider and see if they should be editing or not. That can still be a boolean value.

I think all of this repeated code in all of these modules can actually get moved into the Row Context. We can make the row context react to changes in the parent context. That way all of the Data Modules can stay the same. All a module should care about is if it is editing or not. The logic for controlling that should exist outside of the module.

Remember that we have two nested contexts here. We can do some fancy work inside the nested context to translate that row to know if it should be editing or not.

lordpixel: sure thing, let me work on it.

brianespinosa (changes requested):
- Leave all Data Modules the way they are
- Move logic to determine which row is being edited into RowProvider by looking at TableProvider and using that info to set isEditing in RowProvider.

lordpixel:
> * Leave all Data Modules the way they are
> * Move logic to determine which row is being edited into RowProvider by looking at TableProvider and using that info to set isEditing in RowProvider.

@brianespinosa the changes are applied, thanks for your suggestions, it ended up better that it was.

packages/cascara/src/ui/Table/context/RowProvider.js
```diff
@@ -4,16 +4,24 @@ import { ModuleContext, ModuleProvider } from '../../../modules/context';
...
+    idOfRecordInEditMode,
+    uniqueIdAttribute,
+    ...rest
+  } = grandparentValues;
+  const { record } = value;
+  const recordId = record[uniqueIdAttribute];
+
+  // isEditing is based on wether the record ids are the same
+  const isEditing = recordId === idOfRecordInEditMode;
+
```
brianespinosa, line 17: Oh yes. This is great. Now those Modules don't need to know at all what is happening outside of them. Great use case for separation of concerns. Hoist the isEditing logic up a little higher. <3

packages/cascara/src/modules/ActionEdit/ActionEdit.js
```diff
@@ -17,11 +17,20 @@ const propTypes = {
...
+  const {
+    idOfRecordInEditMode,
+    isEditing,
+    enterEditMode,
+    exitEditMode,
+    formMethods,
+    record,
+    onAction,
+    uniqueIdAttribute,
+  } = useContext(ModuleContext);
```
brianespinosa, line 29: I think in the context of ActionEdit... this might actually be an early indication for us that we will have some Component specific Modules. I THINK this is the right way to do this for Table. But this also means this same Action Module will not work for Form.

We may need to consider making a modules folder inside Table and then calling this something like `TableActionEdit` instead.

Let's leave it here for now but I am pretty sure we made a discovery here about specialization with Action Modules.

brianespinosa (changes requested):
@lordpixel this all looks great now and significantly more simple. 🙌🏽 

Can we also get a few test cases for this to make sure we don't lose the isolation in row editing? After that, I think we can merge.

brianespinosa:
Go ahead and resolve those conversation items when you see them. They are not holding back the PR.

packages/cascara/src/modules/ActionEdit/ActionEdit.js
lordpixel: I thought about this when coding the changes, you are right on the fact that component-specific modules will start to be more evident. This is one of them, and I first realized when I saw some failing tests in another suite (outside Table) and the solution made me think - oh crap this can possibly be a foot-gun - haha.

Let me evaluate the effort for change suggested here, I think it's worth to have good semantics reflected in the file structure as well so if it results to be easy-piece I will make it too.

brianespinosa: @lordpixel if you do change this, I think it would make sense to keep the old ActionEdit but put it inside Form. That is still correct I think for editing a form. For table, the component with changes you made here is correct.

lordpixel: @brianespinosa I am having some problems with the unit test for validating the single-row edit. Can we merge this so FE can have the fix and then push the test case?

brianespinosa: @lordpixel yes. Let's do that.

brianespinosa (approved):
Approving changes. Separate PR will come with unit tests.

@lordpixel please create a separate task to track those unit tests.

PR #139, opened 2021-01-20
FDS-15 Add support for Form to Cascara

brianespinosa:
## New Component Checklist

## To Do
- [ ] Implement the action function for the form
- [ ] Set up the Form-specific FormActionEdit module


brianespinosa:
@lordpixel you can take a look at this now. Form will be a LOT more simple than Table, and it builds on top of some of the good patterns we have there.

This still is missing the ability to submit and the ability to define an edit action module, but otherwise this might be pretty close.

I think we will need to also be able to define FormRow ratios at some point in configuration, but for now I think this is basically what we need.

cc @CaffeinatedAllie

PR #142, opened 2021-02-05
FDS-98 - Disable the CI step that builds the docs

lordpixel:
Hi team,

This disables the CI step that builds the docs.

## New Component Checklist


brianespinosa (approved):
LGTM. Thank you @lordpixel this will make things run a little more smoothly.

PR #143, opened 2021-02-05
feat(FDS-15): Form P1

brianespinosa:
Resolves FDS-15

Checklist:

- [x] make more evident that we can use `placeholders` in text inputs
- [x] make sure `save` button of `edit` module is of `type={'submit}`
- [x] more examples

### Snapshots

- due to additional module updates, some snapshots have changed showing icons
- data-testid attribute was removed from ActionBar, changing the snapshot

### Fixtures
Some fixtures in cosmos have changed

### Dependencies
A new `postinstall` script was added to build Icons


lordpixel (approved):

brianespinosa:
@lordpixel note that this branch is currently building and deploying correctly. So we will need to see what is changed.

It looks like the Omni chat PR is also failing, but not sure if for the same reasons. I am about to pull develop into here to pick up our Epic tag checklist. That will give us another build here to validate if docs are still good in this branch or if something broke in develop too.

brianespinosa:
Looks like there are target branch conflicts. Going to resolve those now before this branch gets borked.

packages/cascara/src/ui/Table/ActionBar.test.snap
```diff
@@ -13,20 +13,22 @@ exports[`ActionBar component tree snapshot 1`] = `
     Espressive
   </span>
   <button
+    aria-label="new"
     class="ui basic button"
     data-testid="new"
```
brianespinosa, line 18: @lordpixel we should remove testid from these.

brianespinosa (commented):
It looks like we are getting pretty close. Just some test-id things that should come out that I suspect are actually not being used in the tests anyway.

packages/cascara/src/ui/Form/Form.js
```diff
@@ -1,12 +1,206 @@
...
+
+// separates the props into two groups:
+//
+// - myProps, all props defined in `propTypes`
+// - otherProps, the rest of the props
+//
+// @param {Object} props, the props object to filter
+// @param {Object} propTypes, a prop-types definition object
+// @returns {object}
+const filterProps = (props, propTypes) => {
```
lordpixel, line 72: change this to `unhandledProps`

brianespinosa:
@lordpixel changing this PR out of draft state and ready to merge. I am marking it to auto merge as soon as you review and approve. Then we should get an alpha release deployed we can test with.

lordpixel (approved):

PR #151, opened 2021-02-13
FDS-29 - Form PAC documentation

lordpixel:
Hi team,

This adds PAC docs for Form. cheers!

### DIFF
--- a/packages/cascara/src/ui/Form/Form.doc.mdx
+++ b/packages/cascara/src/ui/Form/Form.doc.mdx
@@ -1,6 +1,7 @@
 ---
 title: Form
 propTable: Form.js
+status: 'Alpha'
 ---
 
 # Public API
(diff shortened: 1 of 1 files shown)


packages/cascara/src/ui/Form/Form.doc.mdx
```diff
@@ -1,5 +1,282 @@
 ---
 title: Form
-# propTable: Form.js
-status: 'WIP'
+propTable: Form.js
```
brianespinosa, line 3:
```suggestion
propTable: Form.js
status: 'Alpha'
```
I think we should always add an alpha or beta status for new components that might be coming in. We will need to figure out how we decide the difference between these, but since it is new, we should at least add this.

brianespinosa (changes requested):

brianespinosa (approved):

PR #152, opened 2021-02-16
FDS-133 Dangerfile Setup

brianespinosa:
This PR adds support for Danger which we can use to set up some basic checks on our PRs.

For now we are checking to make sure we have assignees and that there is a description. Later on, we can test for some things we need in our PR description based on some other checks on files changed.

### Dependencies
We are updating a couple dependencies in this PR.


brianespinosa:
@lordpixel this is ready to go now. There are not a LOT of rules, but I think there are enough to get us started and we can add from here.

Note I disabled a couple other actions that were previously running, but not really doing what we needed.

lordpixel (approved):
This is great, now that we have these rules, is it just matter of us using [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/)?

PR #154, opened 2021-02-18
FDS-134 - Table action stack (<ActionsMenu />)

lordpixel:
Hi team,

This adds ActionsMenu to Table.

**Closed**
image

**Open**
image

https://projects.invisionapp.com/d/main#/projects/prototypes/21025246


lordpixel:
Snapshot is failing because Reakit generates new WIA/ARIA content on every test
```jsx
<button
      aria-controls="id-4i1qlg"
      aria-controls="id-e15rrn"
      aria-expanded="false"
      aria-haspopup="menu"
      class="icon basic ui button"
      type="button"
>```

File: packages/cascara/src/ui/ActionsMenu/ActionsMenu.js [hunk 1]
```diff
@@ -1,23 +1,18 @@
...
 
+import styles from './ActionsMenu.module.scss';
 import { popperOverTrigger } from '../../shared/popperModifiers';
-
-const DEFAULT_TRIGGER = (
-  <Button className='ui basic icon button'>
-    <b>⋯</b>
-  </Button>
-);
-// const DEFAULT_TRIGGER = <p>'Trigger'</p>;
```
brianespinosa, line 14: @lordpixel we don't want to lose this. We will probably never need to define a different trigger for the ActionsMenu to open than this. But if we do not define this, people will be guessing what was used each time.

The way I was able to see this... if you look at the button in your screenshots, the ellipsis icon is now at the bottom of the button instead of the middle like this character.

Let's bring this back so we do not need to define a trigger when using ActionsMenu.

packages/cascara/src/ui/Table/TableRow.js
```diff
@@ -35,32 +36,48 @@ const propTypes = {
...
+    );
+  };
+
+  const actions = (
+    <td className={styles.CellActions} key={`${id}-actionbar`}>
+      {outsideActions.map(renderActionModule)}
+      {Boolean(insideButtonActions.length) ? (
+        <ActionsMenu
+          actions={insideButtonActions}
+          trigger={<Button className='icon' content='...' />}
```
brianespinosa, line 78: - [ ] Do not define a trigger for ActionsMenu. Instead use the default.

brianespinosa (changes requested):

packages/cascara/src/ui/ActionsMenu/ActionsMenu.js
lordpixel: working on it

brianespinosa (approved):
LGTM

PR #156, opened 2021-02-23
FDS-136 Do not render popover as a modal

brianespinosa:
We should now not have to worry about z-index from parent defined z-index and we will also not have a lot of extra portals sitting in the DOM.

This patch will need to be tested once published to verify.

### Snapshots
Updating Snapshot test due to the Menu no longer rendering in a portal, but instead now rendering adjacent to the menu button


lordpixel (approved):
👍

PR #159, opened 2021-02-24
FDS-138 Return ActionsMenu to portal rendering, and set z-index

brianespinosa:
### Snapshots

Snapshot tests are being updated here because the menu is no longer being rendered inline. There would technically be no way to break out of the positioning here with the popper due to the sticky cell position.

Note that these snapshot tests are not actually showing the menu as it is being rendered in a portal. Something for us to think about in the future.


lordpixel (approved):
👍

PR #163, opened 2021-02-25
chore(modules): take all Test ID code out of our modules and props

lordpixel:
## FDS-135

We are removing `data-testid` from prop type definitions since they are not required. If developers want to add test ids they can still do it manually and leverage the fact that _any unmutted props will be `...spread` into the final component._

> we still need to discuss whether we want to remove the definitions for [ActionEdit](https://github.com/Espressive/cascara/blob/develop/packages/cascara/src/modules/ActionEdit/ActionEdit.js#L10), a special case in which [we do some magic](https://github.com/Espressive/cascara/blob/develop/packages/cascara/src/modules/ActionEdit/ActionEdit.js#L36) to make things easier to test.


brianespinosa (approved):
<3

PR #164, opened 2021-02-26
0.3.3

brianespinosa:
### Dependencies

We added danger.js and incremented versions for our own packages


brianespinosa:
We might want to consider not having our Dangerfile get worried about these PRs to `main` for PR description type stuff if those were already done in PRs to `develop`

brianespinosa:
Closing this PR because it was not approved and merged before other work came in so it is now invalid, which we do not want.

@lordpixel  We need a way to make sure these get approved so they will merge and deploy before other work coming in from develop. Maybe we DO still need to create release branches. We branch from develop when we are going to do a release and submit that release as a PR into `main` that way new things can come into develop and we do not accidentally get things released if another PR gets submitted separately.

PR #168, opened 2021-03-02
chore(fix eslint): Resolve all `import` plugin eslint rules

brianespinosa:
This resolves all lint issues related to the `import` plugin. Half were from files being moved and not correctly defined, which were breaking in Cosmos builds but not failing.

### Dependencies

We updated all package.json files to add a `prepare` lifecycle script for building packages before publishing, etc. This way we can have our CI environment get all packages into a built state if they require it for testing.


lordpixel (approved):
👍

PR #169, opened 2021-03-03
chore(FDS-146): Resolve eslint `prefer-` prefixed errors

brianespinosa:
Resolves FDS-146

## Eslint Rules

```
'prefer-const': 0,
'prefer-destructuring': 0,
'prefer-spread': 0,
'prefer-template': 0,
```


packages/cascara/src/modules/ActionEdit/ActionEdit.js
```diff
@@ -35,9 +35,9 @@ const ActionEdit = ({ dataTestIDs, editLabel = 'Edit' }) => {
   /**
    * this seems like ugly, we need to find a better way
    * to ease testing.. */
-  let cancelTestId = {};
-  let editTestId = {};
-  let saveTestId = {};
+  const cancelTestId = {};
+  const editTestId = {};
+  const saveTestId = {};
```
brianespinosa, line 40: @lordpixel I noticed these during this fix and created a task for removing these test id things here in addition to removing the props which we already did. <3

lordpixel: I added a comment on that task to explain why we need them. We'll need another way of specifying these.

lordpixel (approved):

PR #173, opened 2021-03-03
Feature/fds 152 autosquash

brianespinosa:
## New Component Checklist


brianespinosa:
Moved to an auto-update only action. We may extend with our own ability to label based on issues from that process later on.

PR #174, opened 2021-03-04
chore(FDS-147) - fix eslint 'react/prop-types' and 'react/forbid-prop-types'

lordpixel:
Resolves FDS-147

### Fixes rules
```
'react/prop-types'
'react/forbid-prop-types'
```


.eslintrc.js
```diff
@@ -4,7 +4,7 @@ module.exports = {
   extends: ['@espressive', 'prettier'], // prettier must be last so we override any rules that potentially conflict with Prettier
   overrides: [
     {
-      files: ['**/next-*/**/*', 'docs/**/*'],
+      files: ['**/next-*/**`/*', 'docs/**/*'],
```
brianespinosa, line 7: @lordpixel why was this ` added?

app-tests/_templates/next-template/src/pages/_app.js
```diff
@@ -1,7 +1,15 @@
+import React from 'react';
+import pt from 'prop-types';
+
 import '@espressive/legacy-css';
 
-const MyApp = ({ Component, pageProps }) => {
-  return <Component {...pageProps} />;
+const propTypes = {
+  Component: pt.element,
+  pageProps: pt.shape({}),
```
brianespinosa, line 8: Any time we have a real need for `object` as a rule, we should use it instead of just using `shape` to trick the linter. This would be a good scenario to use `object` and then comment `// eslint-disable-next-line react/forbidden-proptypes -- We do not know what the object params might be in this case`

app-tests/next-button/src/pages/_app.js
```diff
@@ -1,7 +1,15 @@
+import React from 'react';
+import pt from 'prop-types';
+
 import '@espressive/legacy-css';
 
-const MyApp = ({ Component, pageProps }) => {
-  return <Component {...pageProps} />;
+const propTypes = {
+  Component: pt.element,
+  pageProps: pt.shape({}),
```
brianespinosa, line 8: Same with this. Let's actually use `object` and comment why when we disable the rule

.eslintrc.js
brianespinosa: Upon looking at the rest of your fixes... the reason you had to add all the React imports to the Next docs is because your ` here broke the lint rules being disabled for React being in JSX scope. React does not need to be in scope in Next.js. You probably had to fight with the anchor valid rule too because of how Next.js does their links.

This ` needs to be removed on line 7 so these rules go back to being disabled. This also means you will want to go pull all the React imports you added to the Next.js files that are not needed.

app-tests/next-layout-admin/src/pages/_document.js
```diff
@@ -1,3 +1,5 @@
+import React from 'react';
+
```
brianespinosa, line 2: Please go back through all of these next.js files and remove the React imports that were added due to the lint rule being broken in the override for docs.

packages/cascara/src/modules/ModuleError/ModuleError.js
```diff
@@ -4,7 +4,7 @@ import styles from '../DataModule.module.scss';
 
 const propTypes = {
   moduleName: pt.string.isRequired,
-  moduleOptions: pt.array.isRequired,
+  moduleOptions: pt.arrayOf(pt.shape()).isRequired,
```
brianespinosa, line 7: Use `option` when it is valid. 

I wish the lint rule had a way to not let an empty shape pass.

packages/cascara/src/placeholders/AreaPlaceholder/AreaPlaceholder.js
```diff
@@ -3,10 +3,13 @@ import pt from 'prop-types';
 import styles from './AreaPlaceholder.module.scss';
 
 const propTypes = {
+  children: pt.oneOfType([pt.element, pt.arrayOf(pt.element)]),
   /** Color of the border */
   color: pt.string,
   /** Label for the container */
   label: pt.string,
+  /** Label can have style */
+  style: pt.shape({}),
```
brianespinosa, line 12: object

packages/cascara/src/ui/Chat/ChatWidget/TicketResolved.js
```diff
@@ -20,7 +22,9 @@ const SURVEY_OPTIONS = [
   },
 ];
 
-const propTypes = {};
+const propTypes = {
+  value: pt.shape({}),
```
brianespinosa, line 26: object

packages/cascara/src/ui/Dashboard/widgets/WidgetBar.js
```diff
@@ -15,7 +15,7 @@ const propTypes = {
   /** The unique value to index by on `data` */
   indexBy: pt.string,
   /** Values to show from `data` */
-  keys: pt.array,
+  keys: pt.arrayOf(pt.shape({})),
```
brianespinosa, line 18: object

packages/cascara/src/ui/Form/context/FormProvider.js
```diff
@@ -1,7 +1,14 @@
 import React from 'react';
-import { ModuleContext, ModuleProvider } from '../../../modules/context';
+import pt from 'prop-types';
 import { useForm } from 'react-hook-form';
 
+import { ModuleContext, ModuleProvider } from '../../../modules/context';
+
+const propTypes = {
+  children: pt.oneOfType([pt.element, pt.arrayOf(pt.element)]),
+  value: pt.shape({}),
```
brianespinosa, line 9: object

packages/cascara/src/ui/InfoPopover/InfoPopover.js
```diff
@@ -1,12 +1,13 @@
...
 
-import {
-  Popover,
-  // PopoverArrow,
-  PopoverDisclosure,
-  usePopoverState,
-} from 'reakit/Popover';
+const propTypes = {
+  message: pt.string,
+  style: pt.shape(),
```
brianespinosa, line 9: object

packages/cascara/src/ui/Table/Table.js
```diff
@@ -95,7 +95,7 @@ const Table = ({
     );
   }
 
-  const unwantedActionButtonIndex = dataConfig?.actionButtonIndex;
+  const unwantedActionButtonIndex = dataConfig?.actionButtonMenuIndex;
```
brianespinosa, line 98: Was this a bugfix we we missed?

brianespinosa (changes requested):
Looks like we have a ` that accidentally enforced lint rules on the Next.js docs. 😢 That needs to get fixed and then all of the imports of React to add it in jsx scope need to be removed for all docs.

There are a bunch of empty pt.shape definitions that should actually be objects with eslint comments opting out of that rule. I am glad you did this though because now we know to watch for that in FE.

.eslintrc.js
lordpixel: Thanks for catching this.

app-tests/_templates/next-template/src/pages/_app.js
lordpixel: Yes, there were a lot of components with no props at all, I used shape for the sake of speed. We spend too much time trying to figure out what the prop types should be. We should add a rule for preventing us from adding components without prop types.

packages/cascara/src/ui/Table/Table.js
lordpixel: Yes

packages/cascara/src/modules/ModuleError/ModuleError.js
lordpixel: Could you please clarify this request?

app-tests/_templates/next-template/src/pages/_app.js
brianespinosa: Technically you just put that rule into effect with this eslint rule. ;)

packages/cascara/src/modules/ModuleError/ModuleError.js
brianespinosa: If using `pt.object` is valid, we should use that as the type, instead of just using `pt.shape({})` with an empty value.

My comment was just that I wish the forbidden proptypes rule had a way to catch the empty `pt.shape({})` and make it an error. The only request on these is to change them to `pt.object` with a eslint comment to disable the rule explaining why we are doing it.

packages/cascara/src/ui/Table/Table.js
brianespinosa: In the future, let's make sure we are creating a ticket for these when we see them so we know what is going on. If it is one line like this we are okay making the ticket and then including the ticket number in the commit message. We do at least need to track that somewhere.

.eslintrc.js
brianespinosa: @lordpixel I am still seeing the character on line 7 that needs to be removed. Until that character is removed, there are lint rules are not correctly disabled on Next.js projects
```
'jsx-a11y/anchor-is-valid': 0,
'react/react-in-jsx-scope': 0,
```

app-tests/_templates/next-template/src/pages/_app.js
brianespinosa: This empty shape prop was missed.

packages/cascara/src/placeholders/AreaPlaceholder/AreaPlaceholder.js
brianespinosa: This one was missed

app-tests/_templates/next-template/src/pages/index.js
```diff
@@ -1,3 +1,5 @@
+import React from 'react';
```
brianespinosa, line 1:
```suggestion
```

React does not need to be in scope for Next.js

app-tests/_templates/next-template/src/pages/_document.js
```diff
@@ -1,3 +1,5 @@
+import React from 'react';
```
brianespinosa, line 1:
```suggestion
```

React does not need to be in scope for Next.js

brianespinosa (changes requested):
Looks like a couple proptypes were missed and React got added to scope in some of the Next.js files

app-tests/_templates/next-template/src/pages/_app.js
brianespinosa:
```suggestion
  pageProps: pt.object,
```

packages/cascara/src/placeholders/AreaPlaceholder/AreaPlaceholder.js
brianespinosa, line 12:
```suggestion
  style: pt.object,
```

brianespinosa (changes requested):
two left! I think there was a github bug before not showing these in the comments. I had to leave and come back a couple times to get these suggestions to work

brianespinosa (approved):
YES!

brianespinosa:
@lordpixel congrats, we got them all!

I enabled auto merge squash so as long as the checks pass again, since this is approved, it should auto merge for us. <3

PR #176, opened 2021-03-04
chore(FDS-155): React.memo and useCallback example

brianespinosa:
Resolves FDS-155

This fixture should show what happens when we have a handler function that is not wrapped with `useCallback` and how that impacts the use of `React.memo`.

We might actually want to create a collection of fixtures like this to share with the FE team. Hmm…


packages/cascara/src/fixtures/ReactMemo.fixture.js
```diff
@@ -0,0 +1,54 @@
+import React, { useState, useCallback, memo } from 'react';
+
+const Button = ({ onClick, children }) => {
```
lordpixel, line 3: missing prop types

lordpixel (approved):
Nice

PR #178, opened 2021-03-06
feat(FDS-73) Set up @espressive/design-tokens

brianespinosa:
resolves FDS-73, FDS-158

This implements a build process for `@espressive/design-tokens` and creates initial formats to output JSON and CSS variables. We will want to also output SCSS variables to make selecting the CSS variables easier moving forward.

The `style` prop will end up being all of the CSS variables so it can be directly imported.

### Dependencies

This removes `yalc` which is no longer needed for development, and adds `design-dictionary` for use in generating consumable assets from design token JSON files.

## New Component Checklist


lordpixel (approved):
🙌

PR #185, opened 2021-03-11
fix(fds-164): table header not adding an extra column to display acti…

lordpixel:
Fixes FDS-164

### Fixtures

ActionsStack fixture was updated to catch this kind of errors.

### DIFF
--- a/packages/cascara/src/ui/Table/0-Table.mdx
+++ b/packages/cascara/src/ui/Table/0-Table.mdx
@@ -234,26 +234,9 @@ title: Everything together!
 ---
 
 <Table
-  data={[
-    {
-      id: 1,
-      userEmail: 'john@doe.it',
-      userName: 'John Doe',
-    },
-    {
-      id: 2,
-      userEmail: 'jane@doe.it',
-      userName: 'Jane Doe',
-    },
-    {
-      id: 3,
-      userEmail: 'jim@doe.it',
-      userName: 'Jim Doe',
-    },
-  ]}
-  dataConfig={{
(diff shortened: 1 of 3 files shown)


packages/cascara/src/ui/Table/TableHeader.js
```diff
@@ -5,14 +5,17 @@ import ErrorBoundary from '../../shared/ErrorBoundary';
 import { ModuleContext } from '../../modules/context';
 
 const TableHeader = () => {
-  const { dataConfig = {} } = useContext(ModuleContext);
+  // FDS-164: table header not adding an extra column for actions
+  // when new prop actions is passed.
```
brianespinosa, line 9: @lordpixel can we please add a test for this? It's definitely something that is easy to miss (as we did). This can be created as a separate task, but let's definitely get it in.

brianespinosa (approved):

brianespinosa:
Weird. Danger was stuck waiting for status but it had completed. I updated the comment by adding a new line so it would run again.

packages/cascara/src/ui/Table/0-Table.mdx
```diff
@@ -280,6 +263,25 @@ title: Everything together!
         saveLabel: 'Guardar',
       },
     ],
+  }}
+  data={[
+    {
+      id: 1,
+      userEmail: 'john@doe.it',
+      userName: 'John Doe',
+    },
```
brianespinosa, line 272: Let's do this as a test.

PR #187, opened 2021-03-11
Cascara v0.4.1

lordpixel:
Cascara v0.4.1

### Bug Fixes
fds-164: table header not adding an extra column to display actionsMenu (#185) (edd4d60)


lordpixel:
@brianespinosa I think you are the right person to solve those conflicts. I didn't want to venture 'cause I don't know the specifics of `danger.js` impl.

brianespinosa:
@lordpixel please cancel this PR and instead create a `release/0.4.1` branch from `develop` and use that to target `main`.

PR #188, opened 2021-03-11
chore(FDS-163): POC for new app test routes, using BaristaStructure

brianespinosa:
Does not complete the task, but we want to get an actual build of Cascara that this structure can be used from.

### Dependencies

This adds an SVG display dependency to Cascara which we will be leveraging for displaying SVG icons.


lordpixel (approved):
Nice, so, `Layouts` work at the css level, whilst `Structures` work at the component level, correct?

🙌

PR #192, opened 2021-03-12
chore(FDS-148): get all *TestID code out of ActionEdit.js

lordpixel:
Resolves: [FDS-148](https://espressive.atlassian.net/browse/FDS-148)

### Dependencies
We started to use `portals` as part of `ActionsMenu`, our test environment required some improvements. At the same time, we recently made the same changes to `Barista`, matching FE made sense now that they are ahead.

- added `jest-environment-jsdom-sixteen@1.0.3`
- bumped `@testing-library/dom` to `v7.30.0`
- bumped `@testing-library/jest-dom` to `5.11.9`
- bumped `@testing-library/react` to `11.2.5`
- bumped `@testing-library/user-event` to `12.8.3`
- bumped `jest` to `26.6.3`
- bumped `jest-environment-jsdom-sixteen` to `1.0.3`

### Snapshots
Snapshots changed because now we are getting all collumns from the `dataConfig.display` prop, plus one for the `actions`.

### Unit tests
- added unit tests for the current change
- added unit tests for [FDS-164](https://espressive.atlassian.net/browse/FDS-164) (table header not adding an extra column for actions)


packages/cascara/src/ui/Table/Table.test.js
```diff
@@ -332,88 +351,95 @@ describe('Table', () => {
...
+      //     cancelable: true,
+      //   })
+      // );
+      userEvent.click(editButton[0]);
+
+      await screen.findAllByRole('button', {
+        name: 'Cancel',
+      });
+
+      const emailTextbox = screen.getByRole('textbox', { name: 'Email' });
```
lordpixel, line 380: @brianespinosa we are not able to use the `byRole` queryset when testing, this is because all components appear by `role` (textbox) but the `name` resolves to an `empty string`. So, in the test, we cannot specifically-select `email`.

It would be nice if we could add the `aria-label` to all modules, just like we specify `labels` for them. In this case, it would have helped a lot because we can't query by the label.
image

That's why the test are failing. This would be solved if in [DataEmail]() we added the `aria-label` prop... Because `reakit` does not add one.

brianespinosa: @lordpixel are we are already spreading {...rest} props to the modules from the object? Why not just define `data-label` to the object that gets mapped to each module?

brianespinosa: It looks like maybe this comment is old. I looked at the API and based on that the [accessible name](https://www.tpgi.com/what-is-an-accessible-name/) guidelines it would appear that the `label` is what `{ name: X}` would resolve to.

Looks like your tests are updated using that now.

I love that RTL is using accessibility to drive this test selector. <3

brianespinosa: Okay, I see what is going on. The reason `name` is empty is because the input has an `id` and it is linked by a `<label>` that has a `for` prop matching that `id`. In the case we are not passing a label, we are still rendering the `<label>` element in the DOM. There is no value string set inside that so `name` will select as empty.

```
<label
  for="Email"
>
  // This would be where the label string would normally appear which I think would show up with
  // screen.getByRole('textbox', { name: 'Email' })
  <input
    class="Input"
    data-testid="email"
    id="Email"
    name="email"
    type="email"
    value="Hayden.Zieme@espressive.com"
  />
</label>
```

I think this is actually raising an accessibility issue because this is what a screen reader would see for that input and it would be not having a label.

I think you keep the test selector exactly the same. The test can drive our code changes we need to make. The test is correct. The components are failing!

- We should actually NOT render a `<label>` tag if the component is set to `isLabeled: false`. We do still need that node for styling purposes, but I think the CSS will work if we just render a `<span>` instead. (Both label and span have the same default display inline without any CSS).

- We should be setting an `aria-label` on the input with the name instead, but ONLY if the `isLabeled` is set to true. This way we do not render the label in the UI as a tag, but we still label that input directly which will not break screen readers.

brianespinosa (changes requested):
Update our modules for accessibility using the existing selector

packages/cascara/src/ui/Table/Table.test.js
lordpixel: Nice, thanks for the comfirmation, working on the changes. 🙌

lordpixel: @brianespinosa tests are passing now, I added `aria-label` to the `column` [definition](https://github.com/Espressive/cascara/pull/192/files#diff-e57c8635c338b80585061dff8863053d3538c48fba44d62b5594f072d1b58289R84) in `dataConfig.display`.

Since we are ...spreading, the prop reaches the DOM elemen and that makes the test green. This goes to show that although we are not setting those internally, developers can do it if they need to.

I will create an FDS task for implementing the accessibility improvements.

packages/cascara/src/ui/Table/Table.test.snap
```diff
@@ -108,8 +108,10 @@ exports[`Table component tree snapshot test 1`] = `
             >
               <span>
                 <span
+                  aria-label="Email"
                   class="Input"
                   data-testid="email"
+                  label="Email"
```
brianespinosa: @lordpixel I think this is incorrect. `label` is not a valid parameter on a `span` tag.

packages/cascara/src/ui/Table/Table.test.js
```diff
@@ -81,6 +81,7 @@ describe('Table', () => {
           module: 'text',
         },
         {
+          'aria-label': 'Email',
           attribute: 'email',
           'data-testid': 'email',
```
brianespinosa, line 86: IMO we should actually not even be passing `'data-testid': 'email',` in these objects. Instead, we should just be using the selector to check the label.

packages/cascara/src/modules/DataEmail/DataEmail.js
```diff
@@ -23,21 +23,22 @@ const DataEmail = ({
...
-  label,
   value,
   ...rest
 }) => {
   const { isEditing, formMethods } = useContext(ModuleContext);
 
   const renderEditing = (
-    <label htmlFor={label}>
-      {label && isLabeled && <span className={styles.Label}>{label}</span>}
+    <label htmlFor={rest.label}>
```
brianespinosa, line 32: @lordpixel why did we remove label from the object destructuring?

brianespinosa (changes requested):

packages/cascara/src/ui/Table/Table.test.js
lordpixel: it makes sense, let me take them out

packages/cascara/src/modules/DataEmail/DataEmail.js
lordpixel: This is a mistake, this file should have not been committed, let me remove it.

packages/cascara/src/ui/Table/Table.test.snap
lordpixel: done, snapshots have been updated

packages/cascara/src/ui/Table/Table.test.js
lordpixel: [FDS-170](https://espressive.atlassian.net/browse/FDS-170) has been created for this.

package.json
```diff
@@ -23,8 +23,8 @@
     "design-tokens": "yarn workspace @espressive/design-tokens",
     "framer:publish": "dotenv npx framer-cli publish",
     "lint": "eslint ./",
+    "test": "jest -c ./.jest/config.js --detectOpenHandles",
```
brianespinosa, line 26: @lordpixel is it possible to put this `detectOpenHandles` into the config file instead? Usually these types of configurations can be done with the CLI or the configuration file. Since we have a lot in our file, it will probably make things easier for our future selves to see that configuration and not miss that it is happening here.

packages/cascara/src/ui/Table/Table.test.js
```diff
@@ -25,69 +24,63 @@ describe('Table', () => {
   // The test `row actions` corresponds to condition a, whilst the test
   // `editable records` addresses condition b.
   describe('component tree', () => {
-    const whileTheUIisReady = async (miliseconds) =>
-      await new Promise((resolve) => setTimeout(resolve, miliseconds));
     const datasetSize = 100;
```
brianespinosa, line 27: @lordpixel I was trying to take a look at the differences in the snapshot file... and Github cannot load the diff. The snapshot itself is over 30,000 lines! ha

I think this data set is far too large to render the table. I am pretty sure this test in particular runs VERY slow. Also, a data set of 100 records is not how we are using the table as of today. Let's think about a few things here:

- What is this test supposed to help us catch?
- If the test fails, is it going to be easy to tell what went wrong and where?
- Does the test evaluate a real use case?

I think for the first two questions, we do answer that in the comments about the test. Basically this is kind of a smoke test for he table to make sure it will correctly render data based on our API. The Snapshot should tell us if there were markup changes. That makes sense. The name component tree I think also makes sense because this is to make sure it was consistent.

Really this comes down to the last question then. I think 100 records is probably way too large here for this test. Will the test be just as effective rendering 5, 10, 15 records? Probably. I think it will also run much faster and if there are changes in the snapshot, we won't have to look at a diff with 30,000 lines. :)

brianespinosa (changes requested):
If we can move the flag on CLI into the config, that will be good. And I think the snapshot tests should probably be adjusted here for a more performant and real world use case with much fewer lines to render.

I really like your detailed comment about what is in this PR and why we are making the changes. This is such a great example of how to make this more clear and have everyone aware of what is going on and why. I think we should share this as an example with FE. Thanks Manu. Thanks Danger.

packages/cascara/src/ui/Table/Table.test.js
lordpixel: @brianespinosa you are right, in fact, when I work on unit tests for Table I always set the `datasetSize` to 5. I didn't know the snapshots were so large, this definitely has to change! Let me work on it, thanks for the catch!

package.json
lordpixel: We actually need to remove that one, it was useful when fixing unit tests but now it just adds overhead to tests.
`detectOpenHandles` was removed as it is not needed anymore.

packages/cascara/src/ui/Table/Table.test.js
lordpixel: This has been changed to 10, snapshots are smaller now.

brianespinosa (approved):

PR #196, opened 2021-03-24
chore(FDS-179): [Chat, Structures] fix prop type warnings

lordpixel:
Resolves FDS-179.

More prop type fixes, this time for: AdminStructureHeader, ChatAttachment and WidgetBar.

### DIFF
--- a/packages/cascara/src/structures/AdminStructure/AdminStructureHeader.js
+++ b/packages/cascara/src/structures/AdminStructure/AdminStructureHeader.js
@@ -9,7 +9,7 @@ const TestLogo =
 
 const propTypes = {
   logo: pt.node,
-  title: pt.oneOfType([pt.arrayOf(pt.node), pt.node]).isRequired,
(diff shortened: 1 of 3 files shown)


packages/cascara/src/structures/AdminStructure/AdminStructureHeader.js
```diff
@@ -9,7 +9,7 @@ const TestLogo =
 
 const propTypes = {
   logo: pt.node,
-  title: pt.oneOfType([pt.arrayOf(pt.node), pt.node]).isRequired,
+  title: pt.oneOfType([pt.arrayOf(pt.node), pt.node]),
```
brianespinosa, line 12:
```suggestion
  title: pt.oneOfType([pt.arrayOf(pt.node), pt.node]).isRequired,
```
This will definitely be required when using this Header slot component.

brianespinosa (approved):
I made a quick change for the Structure Header slot component that Title will be required when using it. Otherwise, this is good. Will mark to auto squash when tests pass.

PR #197, opened 2021-03-26
chore(FDS-182) Nivo version bump

brianespinosa:
This PR should resolve some lockfile issues that were present and causing issues in our CI install. It also should give us a warning-free installation, which we want to pass along to our consumers.

### Dependencies

- This bumps all Nivo packages to 0.67.0
- We also alias `react-spring` to `@react-spring/web` since Nivo uses it, but the full platform flavor throws warnings about missing platform dependencies, which we do not support. Making sure we are only using the web platform package gets rid of those warnings for our package consumers.

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -8,8 +8,8 @@
     "prepare": "echo 'No prepare scripts for this package'"
   },
   "dependencies": {
-    "@espressive/cascara": "*",
-    "@espressive/legacy-css": "*",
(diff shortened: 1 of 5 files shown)


brianespinosa:
We may need to just stick with `link:../` aliasing for our packages so services like Vercel can deploy without having to try to get packages from our package source.

lordpixel (approved):
🙌

PR #199, opened 2021-04-21
feat(FDS-177) feat(FDS-178) List and Bubble Widget

brianespinosa:
## feat(FDS-178) Widget: Bubble

These are prerelease versions that validate our PAC's before we finalize.

### Dependencies

Updates to a few outdated dependencies with only patch or minor version bumps


lordpixel (approved):
🙌

PR #200, opened 2021-04-27
feat(FDS-62): Admin Structure PAC

brianespinosa:
## Admin Structure Prerelease

Finalizes the API for Admin Structure slots and slot components. There is still work to be done making this fully responsive, but the API of the PAC should not change.

### Dependencies

There are some new dependencies added, as well as some updates to older dependencies to keep things green and address potential security concerns.


lordpixel (approved):
Cant' wait to see this in prod! 🙌

PR #206, opened 2021-05-07
feat(FDS-207): Structure theme object support

brianespinosa:
## Structure Theme Support

There is now a `theme` prop available on `AdminStructure` which accepts an object. Typing is as follows:

```
theme: pt.shape({
    color: pt.shape({
      primary: pt.string,
      secondary: pt.string,
    }),
  }),
```

This is intentionally generic so that we have an object that is flexible enough to extend if we find a need to add further theme support in the future. Make special note that `primary` and `secondary` are not at the root of the theme object but rather they are part of a `color` object like `color.primary` and `color.secondary`.

### Dependencies

We the following two libraries for color processing and checking contrast ratios for accessibility:

- `color`
- `color-string`

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -9,6 +9,7 @@
   },
   "dependencies": {
     "@espressive/cascara": "file:../packages/cascara",
+    "@espressive/design-tokens": "file:../packages/design-tokens",
     "@espressive/legacy-css": "file:../utils/legacy-css",
     "@fluentui/react-northstar": "0.53.0",
     "framer-motion": "2.5.5",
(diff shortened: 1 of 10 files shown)


lordpixel (approved):
This is awesome, the shape of the `color` prop matches the one available as part of the tenant context value. I can't wait to test this together with `esp-auth`. 🙌

brianespinosa:
@lordpixel funny... glad I made the same design choice 4 years ago when we set that up. 

Make sure the internal keys inside the color object are the same. I suspect they are not. The keys for the object in Cascara are probably more generic than the ones being set inside that color object for tenant branding.

PR #212, opened 2021-05-18
feat(FDS-221): Barista theme for Cascara Chat

brianespinosa:
This creates a new Barista theme for Cascara Chat. It is selectable in the Theme Provider playground. We can eventually make it the default for that provider. Note that this should not impact the way Chat works in the Teams app as that application brings its own provider and only switches between the three available teams themes at the app config level.

image


lordpixel (approved):
This looks great 🙌

lordpixel (approved):

PR #216, opened 2021-05-20
chore(FDS-224): Remove usage of SUIR in Cascara and remove it from peer deps

brianespinosa:
This PR removes SUIR from peer dependencies. It also removes some usage of components that were in the codebase and replaces with class lists.

Moving forward we will also want to remove these from fixtures as well.

### Dependencies

We are removing SUIR from peerDependencies


brianespinosa:
Ran `yarn upgrade` and marked snyk as successful

lordpixel (dismissed):

PR #217, opened 2021-05-21
chore(FDS-222): add documentation for modules

lordpixel:
Resolves [FDS-222](https://espressive.atlassian.net/browse/FDS-222).

This adds documentation for `modules` in Cascara Design as well as the playground component.
It also removes a lot of `backtick` enclosed words that distracted the reader more than providing good information.

### Snapshots
Snapshots changed because we deprecated `label` prop in `ActionButton`.

### Fixtures
The new `ModuleSandbox` has been added to the modules fixture


docs/src/components/Code/Code.js
```diff
@@ -13,13 +13,13 @@ theme.plain.backgroundColor = undefined;
...
 const propTypes = {
-  children: pt.oneOfType([pt.element, pt.arrayOf(pt.element)]),
+  children: pt.oneOfType([pt.node, pt.arrayOf(pt.node)]),
   className: pt.string,
   live: pt.bool,
   title: pt.string,
 };
 
-const Code = ({ children, className, live = true, title, ...rest }) => {
+const Code = ({ children = [], className, live = true, title, ...rest }) => {
```
brianespinosa, line 22: I don't think we should be defining an empty array for children as a default. We should instead be checking anywhere we are trying to access children and make sure it is present.

packages/cascara/src/modules/ActionButton/ActionButton.js
```diff
@@ -5,12 +5,14 @@ import { ModuleContext } from '../context';
 import { Button } from 'reakit';
 
 const propTypes = {
-  /** Every action can have a name */
+  /** DEPRECATED - do not use */
   actionName: pt.string,
   /** Presents the button without a label. NOT USER CONFIGURABLE */
```
brianespinosa, line 10: @lordpixel I wonder if we should consider props like this and possibly label them as private. We could then use that information to omit them from the docs. Additionally, we can also then write some functions to allow us to determine which props we want to handle and which are outside of that scope. It will be easier to use functions to pass those handled props where they need to go and let others pass through. This we can also use to protect private props.

brianespinosa:
```suggestion
  /** PRIVATE: Shows a label */
```

packages/cascara/src/modules/DataCheckbox/DataCheckbox.doc.mdx
```diff
@@ -1,10 +1,77 @@
...
+
+Additional to the props specified aside, please use the following property to configure your module:
+
+| property | type                | description                                             |
+| -------- | ------------------- | ------------------------------------------------------- |
+| module   | required **String** | specifies the type of module, in this case **checkbox** |
+
+> Adding this module to your **Cascara** **Table** or **Form** would look like this:
+
+```jsx
```
brianespinosa, line 20: I worry that these JSX files will become a lot to maintain having the Table here. This also obscures the point that these are also used in Form.

I think we should instead render the module itself only in these code examples. If we do it this way... there is a LOT of code we will have to update if we need to change something with Table or Form and we are showing things outside of the scope of the module itself.

If we _did_ want to show this at some point, I think it would make more sense to make a Table and/or Form sandbox just for Docs that has all of this boilerplate inside of it for showing these modules. But I am pretty sure that our lives are going to be a lot easier not maintaining that and instead just render the modules themselves.

packages/cascara/src/ui/Table/0-Table.mdx
```diff
@@ -8,10 +8,9 @@ propTable: Table.js
 ## data
 
 Every object in this array will potencially be rendered as a table row.
-The shape is up to you, the only requirement is that every object (a.k.a. row) must
-have an attribute that uniquely identifies the object in the array.
+The shape is up to you, the only requirement is that every object (AKA row) must have an attribute that uniquely identifies the object in the array.
 
-> We call this attribute `uniqueIdAttribute`, which you will be able to specify later on.
+> We call this attribute **uniqueIdAttribute**, which you will be able to specify later on.
```
brianespinosa, line 13: When referring to a prop or component, we should show it as `code` and not bold.

packages/cascara/src/ui/Table/0-Table.mdx
```diff
@@ -40,23 +39,27 @@ const data = {[
...
 ## actions
 
-Actions represent 'verbs' we want to apply to our Table rows, like 'view' to navigate to another page with detailed information about the row.
-Another example would be 'Delete' to delete a row. 'actions' will be appended to each row, they'll appear as buttons.
-By defining an 'action', you are telling Table how you want to be notified when your 'action' button is pressed. Via the 'onAction' event.
+Think of actions as **verbs** you would want to apply to each of the Table rows, like **view** to navigate to another page with detailed information about the row.
+Another example would be **Delete** to delete a row. actions will be appended to each row as buttons.
 
-> Table does not provide the logic to handle 'actions', the implementation for deleting a row, navigating to another page, etc. it's up to you.
+When you define an action, you are telling Table how you want to be notified when your action's button is clicked. Via the **onAction** event.
```
brianespinosa, line 45: Same here. If `onAction` is a prop, we should show it as code and not bold.

brianespinosa (changes requested):
My general feeling with these modules in docs is that we should be trying to only render the module itself and not as a part of the Table or Form. If there is something preventing us from doing that, I suspect it is because we do not have the correct defaults defined in the module context.

Let's change these JSX examples in the MDX files to render only the module directly and not as part of a Table. If I recall, I think this is actually why we have a separate "private" export so that we can export those privately for use only in the docs.

packages/cascara/src/modules/ActionButton/ActionButton.js
lordpixel: I like the approach you propose @brianespinosa, let's explore how we can tackle this.

The reasonn for this particular prop to be marked as `deprecated` is because we renamed it to just `name`. We still need the old `actionName` at least until deprecation is fully applied.

docs/src/components/Code/Code.js
lordpixel: This was added to make `docs` work at some poinnt,  it will be removed in the next commit.

packages/cascara/src/modules/DataCheckbox/DataCheckbox.doc.mdx
lordpixel: You are right @brianespinosa, let me change the examples to use the module provider

packages/cascara/src/ui/Table/0-Table.mdx
lordpixel: In a review call, you mentioned that it was too much highlighting of words (Since cascara's `code` format is  a little bit outstanding) so I changed it to be just **bold**.  I can change it back to `code`, np.

packages/cascara/src/ui/Table/0-Table.mdx
lordpixel: ok

packages/cascara/src/modules/DataCheckbox/DataCheckbox.doc.mdx
lordpixel: @brianespinosa we are now using both, the `ModuleProvider` and `Table` playgrounds

packages/cascara/src/modules/ActionButton/ActionButton.js
```diff
@@ -20,6 +22,8 @@ const ActionButton = ({
   ...rest
 }) => {
   const { isEditing, onAction, record } = useContext(ModuleContext);
+
+  // @bje we need to decide if we go for content or label here, both makes no sense
```
brianespinosa, line 26: I agree. I was looking at this and thought it was confusing when looking at the comment on the prop type description.

Initially, I was thinking it would make sense for modules to have almost the same shape and you could just switch between them without having to change the shape at all. That might not make sense anymore... at least not for "Action" type modules.

Go ahead and create a separate task for renaming this. We will want to deprecate "label" with a warning for a couple releases, then make sure everything gets migrated in the codebase before we remove that completely.

packages/cascara/src/modules/ActionEdit/ActionEdit.doc.mdx
```diff
@@ -1,11 +1,101 @@
...
+
+    switch (action.name) {
+      case 'edit.start':
+        console.log(`Your action '${action.name}' has been invoked!`);
+        break;
+    }
+  }}
+  // this defines the unique ID field in your data
+  uniqueIdAttribute={'id'}
+/>
```
brianespinosa, line 100: Before we are done, I think all of these examples with Table should come out of the MDX files. This is still a lot of boilerplate to maintain. I assume we are just leaving them here for now while we validate the sandbox example you put together for modules.

packages/cascara/src/modules/DataCheckbox/DataCheckbox.doc.mdx
```diff
@@ -1,10 +1,96 @@
...
+        console.log('Exiting edit mode, edit canceled');
+        break;
+      case 'edit.save':
+        console.log('Exiting edit mode, saving changes');
+        break;
+    }
+  }}
+  // this defines the unique ID field in your data
+  uniqueIdAttribute={'id'}
+/>
```
brianespinosa, line 91: Remove this Table example before we finalize.

packages/cascara/src/modules/DataText/DataText.doc.mdx
```diff
@@ -1,10 +1,93 @@
...
+        console.log('Exiting edit mode, edit canceled');
+        break;
+      case 'edit.save':
+        console.log('Exitin edit mode, saving changes');
+        break;
+    }
+  }}
+  // this defines the unique ID field in your data
+  uniqueIdAttribute={'id'}
+/>
```
brianespinosa, line 90: Remove Table example

packages/cascara/src/modules/DataTextArea/DataTextArea.doc.mdx
```diff
@@ -1,17 +1,98 @@
...
+        console.log('Exiting edit mode, edit canceled');
+        break;
+      case 'edit.save':
+        console.log('Exitin edit mode, saving changes');
+        break;
+    }
+  }}
+  // this defines the unique ID field in your data
+  uniqueIdAttribute={'id'}
+/>
```
brianespinosa, line 95: Remove Table example

packages/cascara/src/ui/Table/0-Table.mdx
brianespinosa: @lordpixel this was marked as resolved, but this still shows up as bold and not as code.

brianespinosa (changes requested):
@lordpixel this looks pretty close. I think your example basically shows that we can easily document a module by itself now so it is safe to remove all of the Table examples from these MDX files now. 🙌🏽 

There are some other requests in terms of renaming the example provider component, moving it into Cascara, exporting it as private, and also using it in a Fixture to show how we can also use it in Cosmos.

Thanks for putting this together. Making ourselves tools like this is going to improve our development experience internally and make working on these complex modules a _lot_ easier.

docs/src/lib/MDX_COMPONENTS.js
```diff
@@ -29,9 +33,89 @@ import {
...
+    </ModuleProvider>
+  );
+};
+
+ModuleExample.propTypes = {
+  children: pt.oneOf([pt.arrayOf(pt.node), pt.node]),
+  isEditing: pt.bool,
+  record: pt.shape({}),
+};
+
```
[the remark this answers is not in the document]
lordpixel: great, I am working on it, adding tasks here:

- [ ] Remove all Table examples
- [ ] Move ModuleExample to Cascara and rename it ModuleExample -> ModulePlayground
- [ ] Add fixture in Cascara to demo how ModulePlayground can be used

docs/src/components/Code/Code.js
```diff
@@ -13,7 +13,7 @@ theme.plain.backgroundColor = undefined;
 theme.plain.fontFamily = undefined;
 
 const propTypes = {
-  children: pt.oneOfType([pt.element, pt.arrayOf(pt.element)]),
+  children: pt.oneOfType([pt.node, pt.arrayOf(pt.node)]),
```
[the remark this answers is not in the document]
lordpixel: Tasks:

- [ ] Create Jira task for `@espressive/prop-types`

packages/cascara/src/modules/ActionButton/ActionButton.js
lordpixel: Tasks:

- [ ] Create a task for deprecating `label` prop in modules, we want to stick to `content`

packages/cascara/src/modules/ActionEdit/ActionEdit.doc.mdx
lordpixel: I wanted to keep Table examples, since that is the way developers are going to use it. The `ModulePlayground` shows  the modules in isolation, but the Table examples show how modules are ultimately used.

- [ ] Remove Table examples from module docs

docs/src/lib/MDX_COMPONENTS.js
```diff
@@ -29,9 +33,89 @@ import {
 
 import { Asciagram, Code, Placeholder } from '../components';
 
+const ModuleExample = ({ children, isEditing, record = {} }) => {
```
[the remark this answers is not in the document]
lordpixel: nice, ModuleSandbox instead of ModulePlayground 👍

docs/src/components/Code/Code.js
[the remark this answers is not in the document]
lordpixel: [FDS-240](https://espressive.atlassian.net/browse/FDS-240) was created to track this work

packages/cascara/src/modules/ActionButton/ActionButton.js
lordpixel: FDS-241 was created to track this request. The changes have been included in this PR.

packages/cascara/src/modules/ActionButton/ActionButton.doc.mdx
```diff
@@ -1,10 +1,92 @@
...
+
+    switch (action.name) {
+      case 'test': // this has to be the same as your action name
+        console.log(`Your action '${action.name}' has been invoked!`);
+        break;
+    }
+  }}
+  // this defines the unique ID field in your data
+  uniqueIdAttribute={'id'}
+/>
```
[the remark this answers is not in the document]
lordpixel: All Table examples have been removed 👍

docs/src/lib/MDX_COMPONENTS.js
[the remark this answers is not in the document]
lordpixel: This has been moved into Cascara modules

packages/cascara/src/ui/Table/0-Table.mdx
lordpixel: my bad, I didn't push my last change, done.

lordpixel:
> @lordpixel this looks pretty close. I think your example basically shows that we can easily document a module by itself now so it is safe to remove all of the Table examples from these MDX files now. 🙌🏽
> 
> There are some other requests in terms of renaming the example provider component, moving it into Cascara, exporting it as private, and also using it in a Fixture to show how we can also use it in Cosmos.
> 
> Thanks for putting this together. Making ourselves tools like this is going to improve our development experience internally and make working on these complex modules a _lot_ easier.

@brianespinosa this is ready for review 👍

packages/cascara/src/modules/DataTextArea/DataTextArea.doc.mdx
lordpixel: channged

brianespinosa (approved):

PR #219, opened 2021-05-31
fix(FDS-226): use commonjs instead of ES6 modules for @espressive/icons

lordpixel:
Resolves [FDS-226](https://espressive.atlassian.net/browse/FDS-226).

The `buildCollection.js` script outputs ES6 modules syntax, which breaks `webpack` build in host app.
This changes the script to output commonjs.


brianespinosa:
I suspect this may need to be resolved in FE. There is something wrong with our Jest config there if it does not know how to deal with a module based package. We SHOULD still do what I said in terms of defining the "module" import for the package instead. I think that is actually the correct solution so that tooling can tell that this is in fact a module only. We don't want to make the whole thing into CJS.

brianespinosa:
Manu, don't get rid of this yet... but I think the updates to Jest should solve this. I am about to make those for us here in Cascara. Then we can circle back. It appears those changes to the Jest config are solving the modules issue in Jest for app_web.

lordpixel:
> I suspect this may need to be resolved in FE. There is something wrong with our Jest config there if it does not know how to deal with a module based package. We SHOULD still do what I said in terms of defining the "module" import for the package instead. I think that is actually the correct solution so that tooling can tell that this is in fact a module only. We don't want to make the whole thing into CJS.

Adding the type `module` line breaks the `docs` build, this is the error message:

````
16:25:24.218 | ReferenceError: require is not defined
-- | --
16:25:24.218 | at file:///vercel/path0/packages/icons/src/buildCollection.js:8:12
16:25:24.218 | at ModuleJob.run (internal/modules/esm/module_job.js:145:37)
16:25:24.219 | at async Loader.import (internal/modules/esm/loader.js:182:24)
16:25:24.219 | at async Object.loadESM (internal/process/esm_loader.js:68:5)
16:25:24.231 | error Command failed with exit code 1.
````

I guess it wannts either ES6 or commonjs, but nnot both.

lordpixel:
adding `"module": "./index.js"` to package.jscon does not resove the issue neither.

brianespinosa:
@lordpixel take a look... but I suspect that this PR can be declined now as we can now support module based packages in Vercel (our docs), and Jest in both our repo as well as the app_web monorepo

lordpixel:
Closing as Jest config did the trick.

PR #221, opened 2021-06-16
feature(FDS-187): 

lordpixel:
## Resolves [FDS-187](https://espressive.atlassian.net/browse/FDS-187).

This updates `Table` so that if either Data nor Colums are defined, it renders nothing.
If either Data nor Colums are defined, it is assumed that we do not  want to display a Table at the moment.

### Fixtures
A new fixture is added `Empty.Table`, to showcase this new behavior.

### Snapshots
A new snapshot test is added to test empty props.


brianespinosa (approved):
LGTM

PR #222, opened 2021-06-21
Feature/fds 245

lordpixel:
### FDS-246
Export prop types for `Table` and `Form` so we can reuse them both inside Cascara, and outside. At this moment we are not exporting form, but the change was made regardless, one thing less to worry about in the future.

### FDS-247
The prop `isEditable` is defaulted inside each of the modules, this is not what we want. The prop must be controlled by the super component that is using the modules, i.e. `Table` and `Form`.

This PR is the first part of two, the other updates our example apps so they reflect this change. Our goal with this change is to not confuse developers by have them define this prop for each of the modules.

We also removed the default value for `isEditable`, `true` was not the correct value, our default should be `falsy`.


packages/cascara/src/modules/DataTextArea/DataTextArea.js
```diff
@@ -22,8 +22,8 @@ const propTypes = {
 
 const DataTextArea = ({
   attribute,
-  isEditable = true,
-  isLabeled = true,
```
brianespinosa, line 26: @lordpixel I think there was a misunderstanding. We do not want to remove these default props. We should keep them. My comment was pointing out that in app_web you are explicitly defining if a module is labeled or not. We should not expose isLabeled outside of Cascara. It should be private.

That means, if someone defines isLabeled on a module object, we should throw it away.

isLabeled is being set by the table or form component. When we map over modules in a Table or Form, we should explicitly set isLabeled for our use case in Table or Form. We should not pass it down from an object config.

There is literally no use case where a Module should be labeled in a Table. Therefore, we should not force developers to keep defining that over and over.

brianespinosa (changes requested):
Put all default props back for isLabeled. We should not have to define that value when putting together a dataConfig. It should not be passed in to a module when rendering a table.

packages/cascara/src/modules/DataTextArea/DataTextArea.js
lordpixel: Changed. @brianespinosa, my reasoning for removing the default `true` in all of those props was to prevent us from unnecessarily passing the prop when is not needed.

A default `true` forces us to pass `false` when we don't actually want a label, it would be easier and less prone to error that if we don't want a label we just simply DO NOT PASS the prop at all.

lordpixel: In other words: `isLabeled` prop should be `false` by default and if the module requires a label we pass `isLabeled` = `true`. Not the other way around.

brianespinosa (approved):
LGTM

PR #223, opened 2021-06-24
chore(FDS-250,FDS-251) Table DX and Cosmos Testing

brianespinosa:
## TableDX

There is a related task with the first items for this work. A test has been created to capture these items. Not all work is completed, so they are currently TODOs in the test.

## Cosmos For Tests

This PR also explores importing Cosmos fixtures which we can use to generate our tests. This pattern could save us a significant amount of effort. We could essentially put together our UI for a test case, then import each fixture into a test. We would have a visual rendering of the actual content for the test to look at while writing tests. This might also be the way we can validate if Cosmos fixtures are still working. If we used a fixture for a test and it gets changed in a way that breaks a test case, we would know the fixture got broken (or potentially the component itself).

### Snapshots

There are new snapshots for the loading and empty states for Table. Additionally, an older snapshot had to be updated because a previous empty state test was removed since it was duplicated in the TableDX tests.


packages/cascara/src/modules/ActionEdit/ActionEdit.js
```diff
@@ -102,44 +102,38 @@ const ActionEdit = ({
 
   return isEditing ? (
     <>
```
brianespinosa, line 104: Removing Tooltip right now as it appears there is a memory leak in the delay on showing the tooltip. We will need to define a cleanup useEffect hook.

packages/cascara/src/modules/DataJson/DataJson.js
```diff
@@ -0,0 +1,72 @@
+import React, { useContext } from 'react';
```
brianespinosa, line 1: See README about this module.

packages/cascara/src/ui/Table/Table.js
```diff
@@ -78,109 +77,35 @@ const propTypes = {
 };
 
 /** This is a Table */
-const Table = ({
```
brianespinosa, line 81: All of this logic got moved to TableBase so that we do not try to run any of it during the initial table render if we have an empty or zero length data.

packages/cascara/src/ui/Table/Table.test.js
```diff
@@ -143,16 +145,6 @@ describe('Table', () => {
...
-    test('snapshot test (empty)', () => {
-      const view = render(
-        <Provider>
-          <Table />
-        </Provider>
-      ).container;
-
-      expect(view).toMatchSnapshot();
-    });
-
```
brianespinosa, line 155: Removed this test since it is now part of the TableDX tests

packages/cascara/src/ui/Table/Table.test.js
```diff
@@ -373,13 +365,13 @@ describe('Table', () => {
...
         screen.getByRole('textbox', {
           container: firstRow,
-          name: 'email',
+          name: 'Email',
         })
       );
       userEvent.type(
         screen.getByRole('textbox', {
           container: firstRow,
-          name: 'email',
```
brianespinosa, line 382: Aria label improvements mean we had to update these

packages/cascara/src/ui/Table/TableBase.js
```diff
@@ -0,0 +1,140 @@
+import React from 'react';
```
brianespinosa, line 1: This is all of the display logic previously in Table so we only run it when we are not trying to render Loading or Empty states

packages/cascara/src/ui/Table/fixtures/ActionStack.Table.fixture.js
```diff
@@ -178,4 +178,4 @@ class Fixture extends PureComponent {
   }
 }
 
-export default Fixture;
+export default <Fixture />;
```
brianespinosa, line 181: Changing these fixtures to JSX elements exposes their props (and state for class components) in Cosmos.

packages/cascara/src/ui/Table/fixtures/TableDX.fixture.js
```diff
@@ -0,0 +1,105 @@
+import React from 'react';
```
brianespinosa, line 1: All of these fixtures here are consumed in the corresponding test below.

packages/cascara/src/ui/Table/fixtures/TableDX.test.js
```diff
@@ -0,0 +1,70 @@
...
+import '@testing-library/jest-dom/extend-expect';
+import cosmosFixtures, { dataResults } from './TableDX.fixture';
+
+const {
+  Loading: loadingFixture,
+  Empty: emptyFixture,
+  'Data Only': dataOnlyFixture,
+  'Data w/ dataDisplay': dataWithDisplayFixture,
+} = cosmosFixtures;
+const RESULTS_KEYS = Object.keys(dataResults[0]);
```
brianespinosa, line 11: This feels kind of gross. It does make our fixtures show nice in the Cosmos UI... but maybe in the long run, maybe we should just label these fixtures with the same camel case fixture naming in the fixture file so it is easier to just import them directly here.

packages/cascara/src/modules/DataJson/DataJson.js
lordpixel: this is awesome!! 🙌

packages/cascara/src/ui/Table/TableEmpty.js
```diff
@@ -0,0 +1,24 @@
...
+  alignItems: 'center',
+  display: 'flex',
+  height: '10em',
+  justifyContent: 'center',
+};
+
+const TableEmpty = () => {
+  return (
+    <div className='ui placeholder segment' style={temporaryEmptyStyles}>
+      <em>No results...</em>
```
lordpixel, line 17: We'll need to add support for i18n.

lordpixel (approved):
this is great! 🙌

packages/cascara/src/ui/Table/fixtures/TableDX.test.js
lordpixel: it's ok, this is way better than the tricky titles I chose, i.e. `EdiTable` 😃

PR #225, opened 2021-07-05
chore(FDS-253): separate out all of the elements of dataConfig and get rid of it

lordpixel:
Resolves FDS-253.

We are moving towards a clean set of props that can be used interchangeably between `Form` and `Table`.
Old props are stripped away and new props are defined and wired-up.

### Dependencies
Index of Cascara was updated to include propTypes exports.

### Fixtures
Form fixture has been updated to include new props.

### Snapshots
The changes in props, plus the fact that we are defining `arial-labels` for `Action` buttons has caused fixtures to change.

### DIFF
--- a/packages/cascara/src/ui/Form/Form.js
+++ b/packages/cascara/src/ui/Form/Form.js
@@ -10,52 +10,83 @@ import { actionModules, dataModules } from '../../modules/ModuleKeys';
 import { formActionModules, formModules } from './modules';
 import ActionBar from './atoms/ActionBar';
 
-const bundledActionModules = {
+// there are two types of actions a form supports:
+//
+// a - actions modules compatible with form
+// b - form-specific modules like submit, reset, clear
+//
+// availableActionModules group them together
+const availableActionModules = {
   ...actionModules,
   ...formActionModules,
 };
 const formDataModules = { ...formModules, ...dataModules };
-const actionModuleOptions = Object.keys(bundledActionModules);
+const actionModuleOptions = Object.keys(availableActionModules);
 const dataModuleOptions = Object.keys(formDataModules);
 
 const propTypes = {
-  // An object of modules to display.
-  //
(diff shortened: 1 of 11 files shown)


packages/cascara/src/ui/Form/Form.js
```diff
@@ -101,18 +132,20 @@ const formFields = (display, data) => {
...
     // In certain predefined-action modules in which a label is not required, e.g. `edit`,
-    // the following unique key generation fails, as it relies on the label (content).
-    const key = `${id}.${module}.${rest.label || module}`;
+    // the following unique key generation fails, as it relies on the label.
+    const key = `${idx}.${module}.${rest.label || module}`;
 
     // TODO: create a function to filter out props that we don't want
+    // eslint-disable-next-line react/forbid-foreign-prop-types -- @brian we need to see if this approach is what we want
+    const { myProps } = filterProps(rest, Action.propTypes);
+
```
lordpixel, line 146: @brianespinosa is [this](https://github.com/Espressive/cascara/pull/225/files#diff-722b073477d424a954017a1cc1c84b8904435755f9099bc392e810383b9eaf7cR63) `filterProps` function what we want?

Perhaps change the shape returned, with a more semantic name..

brianespinosa: Kind of. I think we can rely on prior art here. Take a look at what is happening in the code in libraries like Semantic UI React and also Reakit. There are helper functions to evaluate these slices of props. I think the functions are semantically named something like `handledProps` which makes sense in the context of the component that the library is handling them.

brianespinosa: @lordpixel go ahead and make a separate task to do this research and settle on a pattern. I think that does not need to be part if the early Q3 release so it could go into the Table P3 epic.

This code can stay since it works for now and we can come back to evaluate this when we look at the overall possible hooks based useTableState style API.

PR #230, opened 2021-07-26
chore(FDS-249): all modules need to present the aria-label on the inputs

lordpixel:
Resolves FDS-249.

This is the first step towards having accessible inputs, `aria-label` is added to all inputs.

We still need to discuss our strategy on `display` mode:

- [ ] should we add `aria-label` to spans in `display` mode?
- [ ] is it ok to use both `aria-label` and the `<label>` tag together?
- [ ] if either `aria-label` or `<label>` but not both, which one over the other?

### DIFF
--- a/packages/cascara/src/modules/DataCheckbox/DataCheckbox.js
+++ b/packages/cascara/src/modules/DataCheckbox/DataCheckbox.js
@@ -36,6 +36,7 @@ const DataCheckbox = ({
       <Checkbox
         {...rest}
         {...checkbox}
+        aria-label={label}
         className={styles.Input}
         id={label}
         name={attribute || label}
--- a/packages/cascara/src/modules/DataJson/DataJson.js
+++ b/packages/cascara/src/modules/DataJson/DataJson.js
@@ -36,6 +36,7 @@ const DataJson = ({
       {label && isLabeled && <span className={styles.Label}>{label}</span>}
       <Input
         {...rest}
+        aria-label={label}
         as={TextareaAutosize}
         className={styles.Input}
         defaultValue={jsonValue}
--- a/packages/cascara/src/modules/DataNumber/DataNumber.js
+++ b/packages/cascara/src/modules/DataNumber/DataNumber.js
@@ -34,6 +34,7 @@ const DataNumber = ({
       {label && isLabeled && <span className={styles.LabelText}>{label}</span>}
       <Input
         {...rest}
+        aria-label={label}
         className={styles.Input}
         defaultValue={value}
         id={label}
--- a/packages/cascara/src/modules/DataRadio/DataRadio.js
+++ b/packages/cascara/src/modules/DataRadio/DataRadio.js
@@ -58,7 +58,7 @@ const DataRadio = ({
     <RadioGroup
       {...radio}
       {...rest}
-      aria-label='fruits'
+      aria-label={label}
       className={styles.Radio}
     >
       {options
--- a/packages/cascara/src/modules/DataSelect/DataSelect.js
+++ b/packages/cascara/src/modules/DataSelect/DataSelect.js
@@ -43,6 +43,7 @@ const DataSelect = ({
       {label && isLabeled && <span className={styles.Label}>{label}</span>}
       <Input
         {...rest}
+        aria-label={label}
         as='select'
         className={styles.Input}
         defaultValue={value}
--- a/packages/cascara/src/modules/DataTextArea/DataTextArea.js
+++ b/packages/cascara/src/modules/DataTextArea/DataTextArea.js
@@ -35,6 +35,7 @@ const DataTextArea = ({
       {label && isLabeled && <span className={styles.LabelText}>{label}</span>}
       <Input
         {...rest}
+        aria-label={label}
         as={TextareaAutosize}
         className={styles.Input}
         defaultValue={value}
--- a/yarn.lock
+++ b/yarn.lock
@@ -1636,7 +1636,7 @@
     strip-json-comments "^3.1.1"
 
 "@espressive/cascara@file:packages/cascara":
-  version "0.4.2-alpha.28"
+  version "0.4.2-alpha.31"
   dependencies:
     "@babel/runtime" "7.14.0"
     "@espressive/icons" "^0.0.2-alpha.5"


packages/cascara/src/modules/DataCheckbox/DataCheckbox.js
```diff
@@ -36,6 +36,7 @@ const DataCheckbox = ({
       <Checkbox
         {...rest}
         {...checkbox}
+        aria-label={label}
```
brianespinosa, line 39: We should _not_ be setting `aria-label` on any input that has an associated `<label>` tag that is linked with `htmlFor` to the `id` of the `<input>` itself.

The only time we would want to use `aria-label` on the input is in the rare scenario where we are not defining a label. We should be evaluating this conditionally if isLabeled is being set or not.

brianespinosa (changes requested):
We should build a test that checks a module to see if it has an id assigned and if there is a label with htmlFor assigned. If not, then it should have an aria-label. If there IS a label with htmlFor that matches the id of the input, then there should be NO aria-label on the input since the contents of the label is what will be used by screen readers as the label.

Basically this makes sure we have a single source of truth for what is actually the label or not by only defining it in one way.

brianespinosa:
@lordpixel we could probably TDD this and write a test that makes sure there is only an aria-label if there is no label linked to the input. I have actually been meaning to write a general accessible form test that we could point at form markup to check for this in general. Maybe this is an opportunity to do that.

packages/cascara/src/modules/DataCheckbox/DataCheckbox.js
lordpixel: @brianespinosa we have improved the way we are handling this, `aria-label` is only added when `isLabeled` is `false`.

lordpixel:
> @lordpixel we could probably TDD this and write a test that makes sure there is only an aria-label if there is no label linked to the input. I have actually been meaning to write a general accessible form test that we could point at form markup to check for this in general. Maybe this is an opportunity to do that.

@brianespinosa we now have a unit test to validate this in Table and From, 🙌

packages/cascara/src/modules/DataCheckbox/DataCheckbox.js
```diff
@@ -31,26 +31,39 @@ const DataCheckbox = ({
   const { isEditing, formMethods } = useContext(ModuleContext);
   const checkbox = useCheckboxState({ state: Boolean(value) });
 
+  // do not add aria label if tag is used
+  const ariaLabel = isLabeled ? {} : { 'aria-label': label };
```
brianespinosa, line 35:
```suggestion
  const ariaLabel = isLabeled ? { 'id': label } : { 'aria-label': label };
```
@lordpixel I _think_ in this scenario we can do this and not set the id in cases where an aria label is present, as it is not being used if there is no label tag linking to it.

packages/cascara/src/modules/DataCheckbox/DataCheckbox.js
```diff
@@ -31,26 +31,39 @@ const DataCheckbox = ({
...
+          <span className={styles.LabelText}>{label || attribute}</span>
+        </label>
+      )}
       <Checkbox
+        {...ariaLabel}
         {...rest}
         {...checkbox}
         className={styles.Input}
-        id={label}
+        id={attribute || label}
```
brianespinosa, line 49:
```suggestion

```
I think we can remove this one since we would be conditionally spreading it above from the ariaLabel const

brianespinosa (commented):
Just one quick change for these where we can spread id vs aria-label depending on isLabeled being set

packages/cascara/src/ui/Form/context/FormProvider.js
```diff
@@ -1,3 +1,6 @@
+/* eslint-disable-next-line eslint-comments/disable-enable-pair -- we need this for the one below */
+/* eslint-disable jsx-a11y/no-redundant-roles -- we need this for form to be accessible */
+
```
brianespinosa, line 3: Why are we disabling this lint rule? Some tags have native semantics which do not require redundantly refining the role.

packages/cascara/src/ui/Form/context/FormProvider.js
```diff
@@ -7,11 +10,14 @@ import { ModuleContext, ModuleProvider } from '../../../modules/context';
 
 const propTypes = {
   children: pt.oneOfType([pt.node, pt.arrayOf(pt.node)]),
+  // A form needs to have a name to be accessible
+  // see: https://github.com/testing-library/dom-testing-library/issues/474
+  name: pt.string.isRequired,
```
brianespinosa, line 15: 👍🏽 

Can we please also make a task to address this for Table? The Table also has a form so we need to be consistent with this support.

packages/cascara/src/ui/Form/tests/Form.test.snap
```diff
@@ -52,6 +55,7 @@ exports[`Form snapshot test 1`] = `
           </span>
           <span
             class="Input"
+            id="Last Name"
```
brianespinosa, line 58: This is a use case I do not think we were thinking about yet. For the readonly version, we should not be adding an id parameter to the span. In this scenario, I believe the correct semantics are to set aria-label on the span with the label name as the value, and the innerhtml being the value itself.

brianespinosa (changes requested):
It looks like we have a scenario we need to rethink for the presentation version when we are not editing, which the snapshot is showing us.

There is also an eslint rule being disabled which I do not think we need to be disabling.

packages/cascara/src/ui/Form/context/FormProvider.js
lordpixel: You are correct on the fact that From has an instrinsic role of `form`, But there is a bug in RTL in which Forms do not pop-out when using `ByRole` queries - This was just an attempt to solve the issue by manually adding the role...

packages/cascara/src/ui/Form/context/FormProvider.js
lordpixel: sure thing, working on it.

lordpixel: [FDS-297](https://espressive.atlassian.net/browse/FDS-297) created for this task 👍

packages/cascara/src/ui/Form/context/FormProvider.js
brianespinosa: It seems like [the issue is coming from us not defining a `name` or `aria-labeledby`](https://github.com/testing-library/dom-testing-library/issues/474) on the Form. We should investigate, but this is probably the route we should take instead of disabling the eslint rule. Could you check if adding either of those parameters make the Form show up using that RTL selector?

Worst case scenario though... we almost never want to be disabling an eslint rule for the whole file. We usually only want to be disabling it for next line instead.

lordpixel: This has been removed, we found another way for querying the `form` and focus the test to it.

packages/cascara/src/ui/Form/tests/Form.test.snap
lordpixel: The `...spread` operation was removed from the `readonly` scenario in modules. `aria-label` added.

packages/cascara/src/modules/DataCheckbox/DataCheckbox.js
```diff
@@ -31,26 +32,37 @@ const DataCheckbox = ({
...
         {...rest}
         {...checkbox}
         className={styles.Input}
-        id={label}
         name={attribute || label}
         ref={formMethods?.register}
       />
       {label && isLabeled && <span className={styles.LabelText}>{label}</span>}
-    </label>
+    </>
```
brianespinosa, line 53: I just happened to catch this luckily. For a checkbox, the `<label>` tag is supposed to wrap the checkbox input itself. It looks like that is being changed here.

brianespinosa, line 53: I just happened to catch this luckily. For a checkbox, the <label> tag is supposed to wrap the checkbox input itself. It looks like that is being changed here.

packages/cascara/src/modules/DataEmail/DataEmail.js
```diff
@@ -29,26 +30,35 @@ const DataEmail = ({
...
         className={styles.Input}
         defaultValue={value}
-        id={label}
+        id={attribute || label}
         name={attribute || label}
         ref={formMethods?.register}
         type='email'
       />
-    </label>
+    </>
```
brianespinosa, line 53: It looks like we are changing this markup everywhere and now making the label a sibling of the inputs instead of wrapping them. This has accessibility implications as well as style implications based on how the nesting of the styles are defined. We need to maintain the label tag as wrapping the input.

packages/cascara/src/modules/DataJson/DataJson.js
```diff
@@ -31,25 +32,39 @@ const DataJson = ({
...
         as={TextareaAutosize}
         className={styles.Input}
         defaultValue={jsonValue}
-        id={label}
+        id={attribute || label}
         name={attribute || label}
         ref={formMethods?.register}
       />
-    </label>
+    </>
```
brianespinosa, line 54: Label tag to wrap input

packages/cascara/src/modules/DataNumber/DataNumber.js
```diff
@@ -29,25 +30,34 @@ const DataNumber = ({
...
         className={styles.Input}
         defaultValue={value}
-        id={label}
+        id={attribute || label}
         name={attribute || label}
         ref={formMethods?.register}
         type='number'
       />
-    </label>
+    </>
```
brianespinosa, line 52: label tag to wrap input

packages/cascara/src/modules/DataRadio/DataRadio.js
```diff
@@ -51,16 +60,11 @@ const DataRadio = ({
       {option.label && isLabeled && (
         <span className={styles.LabelText}>{option.label}</span>
       )}
-    </label>
+    </>
```
brianespinosa, line 63: label tag to wrap input

packages/cascara/src/modules/DataSelect/DataSelect.js
```diff
@@ -60,13 +68,15 @@ const DataSelect = ({
           <option value={value}>{value}</option>
         )}
       </Input>
-    </label>
+    </>
```
brianespinosa, line 71: label tag to wrap input

packages/cascara/src/modules/DataText/DataText.js
```diff
@@ -29,26 +30,34 @@ const DataText = ({
...
         className={styles.Input}
         defaultValue={value}
-        id={label}
+        id={attribute || label}
         name={attribute || label}
         ref={formMethods?.register}
         type={'text'}
       />
-    </label>
+    </>
```
brianespinosa, line 52: label tag to wrap input

packages/cascara/src/modules/DataTextArea/DataTextArea.js
```diff
@@ -30,25 +31,34 @@ const DataTextArea = ({
...
         as={TextareaAutosize}
         className={styles.Input}
         defaultValue={value}
-        id={label}
+        id={attribute || label}
         name={attribute || label}
         ref={formMethods?.register}
       />
-    </label>
+    </>
```
brianespinosa, line 53: label tag to wrap input

packages/cascara/src/ui/Form/Form.fixture.js
```diff
@@ -157,10 +161,16 @@ export default {
...
+      name='test-form'
     />
   ),
   dataWithDisplay: (
-    <DataWithDisplay actions={actions} data={data} dataDisplay={dataDisplay} />
+    <DataWithDisplay
+      actions={actions}
+      data={data}
+      dataDisplay={dataDisplay}
+      name='test-form'
```
brianespinosa, line 172: Awesome to see that this apparently works! I did not know about this until I saw the notes in the issue opened in RTL.

brianespinosa (changes requested):
I am not sure how I missed it in the previous review, but we are changing the label tag as no longer being the wrapper around the inputs. This has style implications as these elements are no longer parent-child, but siblings. This is also how we get implicit "for" support on our labels if for some reason we did not get a for and id link between a label and input defined.

This is a risky change to be making. I think we should be maintaining the parent-child relationship. We want to be VERY wary of structural markup changes like this. On Forms they can impact accessibility, but on a larger scale, they can have significant style implications.

packages/cascara/src/modules/DataCheckbox/DataCheckbox.js
lordpixel: Thanks for catching this! The input component must be wrapped by the label, IDK how I missed this - Per our convo in slack, a new `LabeledModule` was created to abstract this logic away from modules.

Modules were cleanned up and code is much easier to follow 👍 
Please let me know if you still see improvement opportunities, this has to be in very good shape for the release.

packages/cascara/src/modules/DataJson/DataJson.js
lordpixel: This module is now wrapped by the new `LabeledModule` 👍

packages/cascara/src/modules/DataNumber/DataNumber.js
lordpixel: This module is now wrapped by the new `LabeledModule` 👍

packages/cascara/src/modules/DataRadio/DataRadio.js
lordpixel: This module is now wrapped by the new `LabeledModule` 👍

packages/cascara/src/modules/DataText/DataText.js
lordpixel: This module is now wrapped by the new `LabeledModule` 👍

packages/cascara/src/modules/DataSelect/DataSelect.js
lordpixel: This module is now wrapped by the new `LabeledModule` 👍

packages/cascara/src/modules/DataEmail/DataEmail.js
lordpixel: Thanks for catching this, we definitely don not want that. This has been solved by the new `LabeledModule` helper component, in which we abstract this logic away and output the module as `children` (wrap).

brianespinosa:
@lordpixel we have conflicts here. Should be good to go after this.

packages/cascara/src/ui/Form/Form.js
```diff
@@ -106,7 +109,9 @@ const formFields = (display, data) => {
     const { myProps: moduleProps } = unhandledProps(rest, Module.propTypes);
 
     return Module ? (
-      <Module {...moduleProps} key={key} label={label} value={moduleValue} />
+      <LabeledModule attribute={attribute} isLabeled label={label}>
+        <Module {...moduleProps} key={key} label={label} value={moduleValue} />{' '}
+      </LabeledModule>
```
brianespinosa, line 114: @lordpixel on one hand, I like that this is more dry because the logic exists in one place, but this logic is part of the module itself, not part of Form. We are placing this wrapper here instead of in the modules themselves. This means every place we ever need to use a module, we would always have to define this wrapper for module logic. This probably significantly breaks the docs because the label logic is now no longer part of the modules and will not even show the label on those modules.

brianespinosa (changes requested):
LabeledModule needs to be part of every module. It is module logic, not Form or Table logic. We cannot work on a module in isolation if part of the logic for the module is not present.

packages/cascara/src/ui/Form/Form.js
brianespinosa: @lordpixel maybe we call this ModuleBase instead of LabeledModule. Any reusable logic for all modules could go on this. Just a thought, but I am pretty sure we will have more boilerplate logic that can go in one place and it does not have to be about only the label.

lordpixel: You actually bring a good point, @brianespinosa, I was thinking the other way around interms of re-usability but I agree this belongs more to a module rather than to the Form.

Working on it, 👍

lordpixel: @brianespinosa the component was renamed to `BaseModule` and the implementation was moved to the modules.

brianespinosa (changes requested):
I think the only thing we may need to do here is add a `...rest` spread on BaseModule so we do not block ourselves from doing any React top level API passing just in case, but other than that, we are probably good and we can do some more of the drying up code in a separate task and separate PR. 

This really does make defining these components much easier now. Definitely something we should consider using a base component in other areas of similar logic. I think we are actually doing something similar for Widgets.

packages/cascara/src/modules/DataCheckbox/DataCheckbox.js
```diff
@@ -2,11 +2,12 @@ import React, { useContext } from 'react';
 import { Checkbox, useCheckboxState } from 'reakit/Checkbox';
 import pt from 'prop-types';
 
+import ErrorBoundary from '../../private/ErrorBoundary';
```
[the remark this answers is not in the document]
lordpixel: working on it.

rociobaena:
Hi @lordpixel :

1. Running All Modules POC section I see the checkbox has a text. 
  [image](https://user-images.githubusercontent.com/87443054/129401940-89203230-2503-461e-ac44-cd486c74bb93.png)
Same for table
image


- I don't know if this was here before or not. I'll make the comment in case we want to create a task for it: In All Modules, without any change in the form, if the Edit button is triggered, this error appears:
image

packages/cascara/src/modules/DataCheckbox/DataCheckbox.js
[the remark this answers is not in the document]
lordpixel: The ErrorBoundary component was moved from modules to BaseModule.

lordpixel:
> Hi @lordpixel :
> 
> 1. Running All Modules POC section I see the checkbox has a text.
>    [image](https://user-images.githubusercontent.com/87443054/129401940-89203230-2503-461e-ac44-cd486c74bb93.png)
>    Same for table
>    image

The Checkbox component had its type to `text`, which is incorrect. Type was set to `checkbox`.

> 
> * I don't know if this was here before or not. I'll make the comment in case we want to create a task for it: In All Modules, without any change in the form, if the Edit button is triggered, this error appears:
>   image

This is an edge case, the fixture is not correctly wrapping the action buttons, so the `enterEditMode`, `exitEditMode` and `onAction` functions do not exists, hence it fails.

packages/cascara/src/modules/DataEmail/DataEmail.js
```diff
@@ -29,38 +31,35 @@ const DataEmail = ({
...
 
   const renderDisplay = (
-    <span>
-      {label && isLabeled && <span className={styles.LabelText}>{label}</span>}
-      <span className={styles.Input} {...rest}>
-        {value}
-      </span>
+    <span {...rest} aria-label={label} className={styles.Input}>
+      {value}
     </span>
```
brianespinosa, line 53: @lordpixel what happened to the input on renderDisplay in this module? It is gone.

packages/cascara/src/modules/DataJson/DataJson.js
```diff
@@ -31,37 +33,39 @@ const DataJson = ({
...
-      <Input
-        {...rest}
-        as={TextareaAutosize}
-        className={styles.Input}
-        defaultValue={jsonValue}
-        id={label}
-        name={attribute || label}
-        ref={formMethods?.register}
-      />
-    </label>
```
brianespinosa, line 46: @lordpixel if we look at the code before and after, and we look at the number of DOM nodes that this code renders before and after, and their parent/child relationships, we are still changing that here.

It is extremely important that we are still rendering the exact same number of nodes with the same structure in both display and editing mode, and we need to be maintaining those with these changes.

packages/cascara/src/modules/DataModule.module.scss
```diff
@@ -16,16 +16,16 @@ $module-background: #f0f0f0;
...
 %ModuleContainer {
   width: 100%;
   padding: $module-padding;
 
   // border-bottom: 1px solid red;
 
-  .LabelText {
-    @extend %ModuleLabel;
-  }
-
```
brianespinosa, line 28: The reason this style is having to be changed is because we are breaking the DOM structure of the modules. The fix is maintaining the same structure. We do not want to be changing the CSS to account for our module markdown changing.

There are other styles that rely on the specific structure of modules being the same that are going to be broken by this change.

Any time we see a need to change CSS that is almost surely a red flag for us that we need to go back and look at our structure.

brianespinosa (changes requested):
@lordpixel unfortunately, this is still not what we need. You addressed the comments in the last review, but the comment from the review before that stating that the structure of Module DOM nodes cannot change (the types can change, but not the count and parent/child relationships) has been forgotten and this PR modifies all of them. There was even a CSS change added to this PR to try to address the styles that broke. 

Please leave this PR in the current state it is in for now. I need to write some tests for all modules that enforce this structure so that we do not run the risk of breaking it. Once those are in, then we can come back to this PR and do some TDD with it using those new tests. If the new tests pass, and we have the logic for showing the labels correctly, then we will be safe to merge. I think using the tests is the best way to make this clear and enforce nothing gets broken.

I am going to mark this PR as DNM while we wait for those tests to get written. I may get to them today.

PR #231, opened 2021-07-27
test(Cascara): [FDS-194] test optimization

lordpixel:
Resolves FDS-194.

### Tests optimization
We are adopting the pattern shown in Pagination, in which we leverage `cosmos` fixtures for testing purposes. This splits the tests found in the legacy `Table.test.js` file into multiple files and moves everything to the `tests` folder.

#### Findings
- [using](https://github.com/Espressive/cascara/pull/231/files#diff-73ad3b1cd75895aee157632078986b9c6a87f0098558598eb6490f5b80f35ef6R37) `within` for snapshot tests can be used to shrink snapshots
- splitting tests helps suite to run faster
- individual tests files run slower, some of them are marked in red by Jest
- using `Provider` from `reakit` seems to be slowdowing tests down
- using `Cosmos Fixtures` seems to to be slowdowing tests down

### Snapshots
Snapshots changed due to test optimization.

### Fixtures
- Snapshot tests
- UX fixture and tests created to encapsulate user-facing scenarios
- DX fixture and tests updated to add tests moved from legacy `Table.test.js` file


lordpixel:
plucking `snapshots` out to their own file seems to only reduce _less_ running time, hence we went for the `dataset` size reduction:
image

lordpixel:
Which compared to [historical running times](https://app.circleci.com/pipelines/github/Espressive/cascara/1647/workflows/aa923583-b085-4ea6-b421-f9680dfad863/jobs/6085) is not that bad.

packages/cascara/src/ui/Table/fixtures/TableDX.test.js
```diff
@@ -9,6 +9,7 @@ describe('Table DX', () => {
   test(`should show a loading state if the 'data' prop is 'undefined' or 'null'`, () => {
     const { baseElement } = render(loading);
     // This technically captures the entire fixture in a snapshot. We should consider if we want to only select the element within for a snapshot or not.
+    // update: by default, snapshots are wrapped by a `<body>` tag, which doesn't cause snapshots to grow signigicantly
```
brianespinosa, line 12: @lordpixel note that the reason I am pointing this out in the DX test is that if someone needs to change the copy in the fixture, the snapshot would technically fail and need to be redone. This is why I am thinking we might want to explicitly use the RTL `within` selector to find our specific component in the fixture and base our snapshots off that only.

For now this is not happening so if the description inside the Fixture changes, the snapshot will need to get updated. Maybe that is okay for now, but it is something to be aware of.

packages/cascara/src/lib/mock/fakeData.js
```diff
@@ -15,11 +15,6 @@ export const generateFakeEmployees = (qty) =>
       country: faker.random.arrayElement(['USA', 'Brazil', 'Argentina']),
       eid: faker.datatype.uuid(),
       email: faker.internet.email(firstName, lastName, 'espressive.com'),
-      employeeNumber: faker.datatype.number(),
-      fullName: `${firstName} ${lastName}`,
-      homePhone: faker.phone.phoneNumber(),
-      officePhone: faker.phone.phoneNumber(),
-      title: faker.name.jobTitle(),
```
brianespinosa, line 22: @lordpixel this looks like a good optimization and something to think about for future tests.

This has me wondering... how long would the tests take to run on average if this data was just JSON and not being generated from a JS library? I think we should consider that. If static data is more performant, we might want to say we limit our actual tests to static data only and only use generated data in development. I don't know the answer to this either way.

packages/cascara/src/ui/Table/Table.test.js
```diff
@@ -57,86 +57,54 @@ describe('Table', () => {
...
+          },
+          {
+            label: 'USA',
+            value: 'USA',
+          },
+        ],
+      },
+    ];
 
     test('snapshot test', () => {
```
brianespinosa, line 101: @lordpixel it looks like you did a test separating out the snapshot from the rest of this test and that did decrease the overall runtime and we went with reducing the test data size to make the tests run faster instead because that gave us a larger margin of improvement.

Did we evaluate what doing _both_ of these things does to the time? Do we get slightly better performance still from both separating tests and also reducing the data test size?

brianespinosa (commented):
I left some comments here on files but I am not requesting any specific changes unless those questions or comments will surface even more performance improvements.

I see included in this info about the overall test time, but I would like us to also look at how long each test itself takes to run. Jest does a good job of indicating when we have test files that take a long time and they even indicate the long ones with a red time telling us the test runs long (we have a LOT of those in the monorepo).

It might be a good idea to establish what we feel our test best practices are to keep tests running fast, make them easier to maintain, etc. These are just guess examples and I do not know if they will actually be true, but maybe something like: 
- Break tests into separate files based on the types of functionality (loading/empty, warning/error, snapshots, display) 
- Keep data set sizes small (but large enough to evaluate all use cases)
- Leverage modern RTL methods like render/screen instead of importing lots of selector methods

packages/cascara/src/ui/Table/Table.test.js
lordpixel: Yes we did, with both approaches, we observed is an increase in running time, not a decrease. It's the last screenshot I posted.

packages/cascara/src/ui/Table/fixtures/TableDX.test.js
lordpixel: @brianespinosa I thought about this when we started to use fixtures for unit tests... It seems to me like we are bloating the tests and then wanting for them to run blazing fast... Aren't we shooting ourselves in the foot with this?

packages/cascara/src/lib/mock/fakeData.js
lordpixel: I didn't think about removing `faker` from the equation, but then again, why do we use these kind of tools if at the end we are going to remove them just because tests take 3 seconds more to run?

Shouldn't we think twice before adding libraries, fixtures and all these things?

brianespinosa: Faker is a helpful tool if you do not yet know what data to use. Or if you have a use case where you need a LOT of data. It is just a tool like anything else. Part of our job is figuring out which tool to use in each use case. I think Faker still has use cases for Cascara repo, but it might not be good specifically for tests if we find out it is not as performant as we want.

packages/cascara/src/ui/Table/fixtures/TableDX.test.js
brianespinosa: @lordpixel what are you saying is bloating the tests? The use of `within` from RTL, or the fact that `render` might contain an h2 or p tag with description text?

brianespinosa (commented):
@lordpixel could you summarize your findings on these:

- Did you determine that the tests run more efficiently if we use static JSON instead of Faker? If so, what sort of difference are we looking at?
- I see the snapshot test being split now. My guess is that the tests for the whole project might not run faster, but individual test files will run faster. Did you observe this and is that why you decided to split the TableSnapshot?

lordpixel:
> @lordpixel could you summarize your findings on these:
> 
> * Did you determine that the tests run more efficiently if we use static JSON instead of Faker? If so, what sort of difference are we looking at?
> * I see the snapshot test being split now. My guess is that the tests for the whole project might not run faster, but individual test files will run faster. Did you observe this and is that why you decided to split the TableSnapshot?

I am working on these, I lost the track because tests for pagination were added and running time increased, I'll post my findings

lordpixel:
Addressing questions:

#### Did you determine that the tests run more efficiently if we use static JSON instead of Faker? If so, what sort of difference are we looking at? 
We are using [entities](https://github.com/Espressive/cascara/pull/231/files#diff-6d5f38ffb0b4002471dc1adc0853d13d5fd88209bfc658f1524d72aebd36dc4d) for tests instead of `faker`, it's a response from the real API. Tests run faster now, but that was not the only change so I cannot give an accurate measure.

#### I see the snapshot test being split now. My guess is that the tests for the whole project might not run faster, but individual test files will run faster. Did you observe this and is that why you decided to split the TableSnapshot?
Tests do run faster, e.g. in `watch` mode they take 5 seconds, but when ran with `yarn test` they take 9 seconds - but this isn't always the case, as we spoke in our call, running time varies depending on the CPU loads.

packages/cascara/src/ui/Table/fixtures/TableDX.test.js
lordpixel: @brianespinosa what I am saying is that things like `faker` or `cosmos fixtures` might be overkill for tests and are just adding even more load, hence tests run sluggier.

brianespinosa (approved):

PR #232, opened 2021-07-27
doc(Cascara): [FDS-248] Update docs to match P2 changes

lordpixel:
Resolves FDS-248.

### Documentation
Cascara docs are oudated and do not reflect important changes generated in P1 and P2, this updated docs to match the most recent alpha, looking forward for the next graduated release.

- Cascara docs updated to match P2 changes

### Fixtures
The `isLabeled` prop was removed from docs as it won't be exposed to developers


packages/cascara/src/ui/Table/0-Table.mdx
```diff
@@ -278,31 +296,29 @@ With the four props described earlier, we can easily create a table: (open your
...
-        label: 'Name',
-        module: 'text',
-      },
-    ],
-  }}
+  dataDisplay={[
+    {
+      attribute: 'id',
+      isEditable: false,
+      isLabeled: false,
```
brianespinosa, line 303: We not define isLabeled anywhere in our dataDisplay objects.

Could you do a search and find any other places that this is being done to make absolutely sure we do not do this anymore? I keep seeing it so I am worried that this is coming from copy paste, which means developers are going to do it too.

brianespinosa (changes requested):
Do not set `isLabeled` on dataDisplay objects

packages/cascara/src/ui/Table/0-Table.mdx
lordpixel: The `isEditable` prop was removed from docs. A quick search didnt throw any other unwanted appeareances.

lordpixel: I totally misread the prop name 🤣 , `isLabeled` was removed from docs.

packages/cascara/src/modules/DataTextArea/DataTextArea.doc.mdx
```diff
@@ -24,7 +24,6 @@ Feel free to play around with the following code:
   <DataTextArea
     attribute='firstName'
     isEditable={true}
-    isLabeled={true}
```
brianespinosa, line 27: @lordpixel in the scenario of documentation if we are showing what is possible on a module, setting a prop on a component should be okay, even if that prop is something that we do not let end-users set.

The problem area is if we are trying to set isLabeled in a module _object_. Those are the ones we should get away from.

@CaffeinatedAllie just a heads up that the actual docs might not show a label for modules but the label will indeed be set correctly from the 'label' prop.

brianespinosa (approved):

PR #233, opened 2021-07-27
chore(FDS-274) directory cleanup

brianespinosa:
This does a lot of work cleaning up our directories a bit. I did not touch Table yet since @lordpixel you have some incoming PRs so we can wait for those to come first and clean them to match after.

This makes folders with complex components easier to determine. It also makes Cosmos directories cleaner and it is easier to see which components are getting exported as public vs private. We might even be able to make setting that up easier now that we have them in separate directories.

### Snapshots

Snapshots should not have changed from this, but they were moved into a /tests directory with their associated tests


lordpixel (dismissed):
🙌

brianespinosa:
@lordpixel conflicts resolved. I think this one will be ready to come in too once changes reviewed.

lordpixel (approved):

PR #234, opened 2021-07-27
chore(FDS-286) Move Pagination into Cascara

brianespinosa:
This also cleans up how we are defining index exports to make it easier for ourselves. There may be conflicts between this and the other PR that cleans up our folder structure. I will resolve them if that happens, whichever PR makes it in first.

@Espressive/design-engineering note that there is some change to how we are defining index.js exports so they are easier to maintain. With the wildcard exports in the private package, that will give us anything in Cascara we need to do POC work if we need something from lib, hooks, etc.

@lordpixel this change might have impacted some of the MDX files for the docs in Table due to local and global modules having the same name. I renamed the Table version with a unique name. We will see how that impacts the docs build.

## New Component Checklist

- [x ] Includes unit tests for _ALL_ required FDS use cases
- [ x] Does not mute any top level API props in React
- [x ] Includes an explanation for any changes to Jest snapshots (should be a PR tag automatically added if this happens)

### Dependencies

We had to add a nextjs plugin to make sure we can transpile modules

## Snapshots

We are adding snapshots for Pagination 🙌🏽


brianespinosa:
Well, that failed docs build answers my question. I will take a look at this in about 60 min.

lordpixel (approved):
🙌

PR #236, opened 2021-07-29
chore(FDS-255): Add empty and loading status for a Form

rociobaena:
## New Component Checklist

-- No unit test for now until we agree with this implementation.
Form is different than Table due to data is expected to be present or not. The form could not have default values to fill the HTML form. For this reason, I added the isLoading property to differentiate between both states.

### Snapshots
Snapshots updated to reflect loading and empty status.

image
image
image
image
image


packages/cascara/src/ui/Form/Form.js
```diff
@@ -9,6 +9,9 @@ import ModuleError from '../../modules/ModuleError';
 import { actionModules, dataModules } from '../../modules/ModuleKeys';
 import { formActionModules, formModules } from './modules';
 import ActionBar from './components/ActionBar';
+import FormLoading from './FormLoading';
+import FormEmpty from './FormEmpty';
```
brianespinosa, line 13: Can we rename these to TemporaryLoading, TemporaryEmpty and set these on both Form and replace the ones on Table? 🙏🏽

packages/cascara/src/ui/Form/Form.js
```diff
@@ -170,11 +176,18 @@ const Form = ({
...
-  const [isEditing, setIsEtiding] = useState(() =>
-    !isEditable ? false : isInitialEditing
-  );
+  const [isEditing, setIsEtiding] = useState(() => {
+    const isDataEmpty = isEmpty(data);
+    if (isDataEmpty) {
+      return true;
+    }
+    return !isEditable ? false : isInitialEditing;
+  });
```
brianespinosa, line 185: Let's explore setting a message with useDeveloperMessage if `isLoading && !isInitialEditing` as this is an invalid state.

packages/cascara/src/ui/Form/Form.js
```diff
@@ -162,6 +167,7 @@ const Form = ({
   onAction,
   isEditable: incomingIsEditable,
   isInitialEditing = false,
+  isLoading = false,
```
brianespinosa, line 170: For isLoading we can try using the getStateFromData function to infer isLoading and isEmpty. That function however probably needs to get updated to work with objects as I think it is only working with arrays.

const { isLoading } = getStateFromData(data)

packages/cascara/src/ui/Form/Form.js
```diff
@@ -185,20 +198,29 @@ const Form = ({
...
+          }}
+          {...rest}
+        >
+          {isLoading && <FormLoading />}
+          {!isLoading && (
+            <>
+              {fields}
+              <ActionBar actions={renderedActions} />
+            </>
+          )}
```
brianespinosa, line 219: Take a look at Table. I think we can probably just do isLoading ? TempLoading : FormContent since they can never show at the same time.

packages/cascara/src/ui/Form/FormEmpty.js
```diff
@@ -0,0 +1,20 @@
...
+
+const FormEmpty = () => (
+  <div className='ui placeholder segment' style={temporaryEmptyStyles}>
+    <em>No data...</em>
+  </div>
+);
+
+FormEmpty.propTypes = propTypes;
+
+export default FormEmpty;
```
brianespinosa, line 20: When these get renamed, I think we can place them inside /private so everyone can use them and we can make sure we are not exporting

brianespinosa (changes requested):
A few changes but generally right on the money for Cascara patterns, and you are even optimizing some of the JSX code that we had not optimized yet. 🙌🏽

brianespinosa:
@rociobaena please also change the title of the PR to follow conventional commits

packages/cascara/src/ui/Form/Form.fixture.js
```diff
@@ -122,15 +116,44 @@ const BasicForm = (props) => {
...
+    <Form {...props} />
+  </>
+);
 export default {
-  basic: <BasicForm actions={actions} data={data} dataDisplay={dataDisplay} />,
+  dataOnly: <BasicForm actions={actions} data={{}} dataDisplay={dataDisplay} />,
+  dataWithDisplay: (
+    <DataWithDisplay actions={actions} data={data} dataDisplay={dataDisplay} />
+  ),
+  empty: <EmptyForm actions={actions} data={data} dataDisplay={[]} isLoading />,
```
brianespinosa, line 156:
```suggestion
  initialEditing: <EmptyForm actions={actions} data={data} dataDisplay={[]} isInitialEditing />,
```

Technically, I think this fixture will change to use isInitialEditing. We probably want to rename it for initial editing as well.

We may also want to add one more fixture for showing data empty (isLoading) and not setting isInitialEditing and make sure it shows the warning we expect.

packages/cascara/src/ui/Form/Form.js
lordpixel: @brianespinosa IDK if we moved the `getStatusFromDataLength` util from the monorepo, just in case here it is so @rociobaena can use it in Cascara:

```javascript
const getStatusFromDataLength = (totalRecordCount) => {
  const isLoading = !totalRecordCount && totalRecordCount !== 0;
  const isEmpty = !isLoading && totalRecordCount === 0;

  return {
    isEmpty,
    isLoading,
  };
};

export default getStatusFromDataLength;
```

lordpixel (approved):
This looks great @rociobaena! I am approving, looking forward for the changes requested by @brianespinosa  🙌

lordpixel:
@rociobaena we can have a call to get this PR ready if you want, let me know 😃

packages/cascara/src/ui/Form/Form.js
brianespinosa: @rociobaena I think the function @lordpixel is mentioning above was actually moved already because Pagination is officially in Cascara now. ;)

packages/cascara/src/ui/Form/Form.js
rociobaena: New files created

packages/cascara/src/ui/Form/Form.js
rociobaena: Using getStatusFromDataLength instead.

packages/cascara/src/ui/Form/Form.fixture.js
rociobaena: Status added:
dataOnly, dataWithDisplay,empty, initialEditing,loading

packages/cascara/src/ui/Form/FormEmpty.js
rociobaena: File removed

packages/cascara/src/ui/Form/Form.js
rociobaena: File global for Form is added, just like it is in Pagination.

packages/cascara/src/ui/Form/Form.js
```diff
@@ -178,6 +180,10 @@ const Form = ({
     !isEditable ? false : isInitialEditing
   );
 
+  const isInvalidFormState = !dataDisplay && !isInitialEditing;
```
brianespinosa, line 183: @rociobaena I think we could make some of these names more clear. The semantics of our name here should make sense when we use it below in the developer message warning. Why is the state invalid? We do not know until we go back and look at where the const is defined. Maybe something like this with also a comment would be more clear?
```suggestion
  // A form cannot be in an initial editing state while also not having a dataDisplay defined
  const isEditingWithoutDataDisplay = !dataDisplay && !isInitialEditing;
```

packages/cascara/src/ui/Form/Form.js
```diff
@@ -178,6 +180,10 @@ const Form = ({
...
+
+  useDeveloperMessage(isInvalidFormState, WARNING_STRINGS.MISSING_STATE_FORM);
```
brianespinosa, line 185:
```suggestion
  useDeveloperMessage(isEditingWithoutDataDisplay, WARNING_STRINGS.INVALID_EDITING_AND_DISPLAY);
```
I suspect this was just from copy paste. :) The warning key should be descriptive of what the state is. It still is talking about missing state... but the warning is actually not about state hook missing. It is actually about an invalid combination of props. Perhaps `MISSING_STATE_FORM` could be something like `INVALID_EDITING_AND_DISPLAY`?

packages/cascara/src/ui/Form/__globals.js
```diff
@@ -0,0 +1,5 @@
+const WARNING_STRINGS = {
+  MISSING_STATE_FORM: 'Form requires dataDisplay to present a form component.',
```
brianespinosa, line 2:
```suggestion
  INVALID_EDITING_AND_DISPLAY: 'Form requires a dataDisplay prop in order to also show isInitialEditing',
```

Since this message is for developers, it is okay for us to reference specific props and talk about how they should be used in combination

packages/cascara/src/ui/Pagination/Pagination.fixture.js
```diff
@@ -54,7 +54,7 @@ const WithStateHook = (fixtureProps) => {
       </p>
       <Pagination {...fixtureProps} state={paginationState} />
       <Pagination {...fixtureProps} state={paginationState} />
-      <JsonPlaceholder data={paginationState} isInitialOpen title='state' />
```
brianespinosa, line 57: I don't think we want to be commenting these out. Do do we recall the reason?

brianespinosa (changes requested):
Few small changes and this should be ready to come in. 🙌🏽

packages/cascara/src/ui/Form/Form.js
rociobaena: Agree on @brianespinosa . Adding the comment.

packages/cascara/src/ui/Form/Form.js
rociobaena: Yeah. I was thinking more in continue to follow the idea from the other component. I will change it.

packages/cascara/src/ui/Pagination/Pagination.fixture.js
rociobaena: Thank you so much for this comment. I commented just for a moment because of the error in eslint and then, I forgot to recover it back.

brianespinosa (approved):

brianespinosa (approved):
@rociobaena These changes look good. Take a look at the list of build tests. The CircleCI tests are now passing. The issue is two other tests still failing. We do not _ONLY_ have CircleCI tests on this repo.

brianespinosa (dismissed):

PR #240, opened 2021-08-12
fix(FDS-298) fix vercel

brianespinosa:
This also officially moves us to yarn v3

Due to the updates to the docs, we also saw a number of issues that come up that always existed but were hidden to us until now. We likely have more issues to solve in the docs as well where we are passing invalid prop shapes, or we have bad prop types.

### Dependencies

We have to move some dependencies around for following correct practices for dep declarations


.circleci/config.yml
```diff
@@ -81,7 +81,7 @@ jobs:
           pkg-manager: yarn
       - run:
           name: Prepare Packages
-          command: yarn workspaces run prepare
+          command: yarn workspaces foreach run prepare
```
lordpixel, line 84: 👍

lordpixel (approved):
PnP FTW 👍 !!

brianespinosa:
> PnP FTW 👍 !!

No PnP yet. We are still using node_modules dirs for now, but we do have zero installs running now. We will still have some work to do in order to get PnP working.

PR #249, opened 2021-09-13
chore(FDS-272): [@espressive/eslint-config] - relax no-console rule

lordpixel:
## ESLint rule `no-console`
- Disable ESLint `no-console` rule in `development` environment.


utils/eslint-config/src/rules/base.js
```diff
@@ -23,7 +23,7 @@ module.exports = {
     eqeqeq: 'error',
     'multiline-comment-style': ['error', 'separate-lines'],
     'no-alert': 'error',
-    'no-console': 'error',
```
brianespinosa, line 26: @lordpixel I think this is fine for now.

I suspect that we will need to actually make more files with configs per environment that we will export entire objects from. We might not get there until next quarter though.

brianespinosa (approved):

utils/eslint-config/src/rules/base.js
lordpixel: ok, I'll dive in to the configuration for ESLint a bit more as a step-ahead. 👍

PR #251, opened 2021-09-13
chore(FDS-299): Setup Commitizen

lordpixel:
## Setup commitizen

Resolves [FDS-299](https://espressive.atlassian.net/browse/FDS-299).

Our repository now follows the [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/).
image

### Dependencies

- Adds `commitizen` at `v4.2.4`
- Adds `cz-conventional-changelog` at `v3.3.0`

### Hooks
- updates Husky's pre-commit hook to run commitizen if all check passed

### Docs
- Adds contributing guidelines
conventional


brianespinosa (approved):
This appears to be good. We might want to consider changing to a different conventional adapter that helps with Jira ticket numbers, but I think we are good for now.

PR #252, opened 2021-09-13
feat(FDS-172) Boundaries w/ Suspense & AdminStructure suspending out of the box

brianespinosa:
This PR further extends the Boundaries component to also add a Suspense boundary with it. It comes with sane defaults, but can also be extended with a React component if a different suspense boundary should show.

Also included: 
- Improves typing for Boundaries to more strictly check if we have an actual React component with a custom prop type
- Adds boundaries to Widgets in Dashboards so that if an individual widget blows up, it is limited only to that widget
- Adds boundaries to Tabs, which are currently displaying content that can be lazy loaded
- Replaces the old ErrorBoundary component in a couple safe places to do so... a separate task for moving the rest of our use cases will be done later.

### Dependencies

Adds `react-is` for doing more strict evaluation to determine if we have an actual component being passed to our error boundaries.


lordpixel (approved):
🙌

packages/cascara/src/modules/ModuleErrorBoundary/ModuleErrorBoundary.js
```diff
@@ -10,9 +10,7 @@ const propTypes = {
 
 const ModuleErrorBoundary = ({ children }) => {
   return (
-    <Boundaries ErrorFallbackComponent={ModuleErrorFallback}>
-      {children}
-    </Boundaries>
+    <Boundaries ErrorComponent={ModuleErrorFallback}>{children}</Boundaries>
```
brianespinosa, line 13: @lordpixel I did a quick rename once I added support for Suspense in our Boundaries component so we would not have overly complex names.

packages/cascara/src/structures/AdminStructure/AdminStructure.js
```diff
@@ -30,9 +31,15 @@ const AdminStructure = ({ children, header, nav, theme = THEME_DEFAULTS }) => {
...
-      <AdminStructureChildren>{children}</AdminStructureChildren>
+      <header className={styles._header}>
+        <Boundaries>{header}</Boundaries>
+      </header>
+      <nav className={styles._nav}>
+        <Boundaries>{nav}</Boundaries>
+      </nav>
+      <AdminStructureChildren>
+        <Boundaries>{children}</Boundaries>
+      </AdminStructureChildren>
```
brianespinosa, line 42: Every slot on AdminStructure now has a wrapping boundary which will support both Suspense and error catching.

packages/cascara/src/system-components/Boundaries/Boundaries.js
```diff
@@ -1,19 +1,53 @@
...
+// This will need to become a custom prop type we can use elsewhere
+const isReactComponent = (props, propName, componentName) => {
+  if (props[propName] && !isValidElementType(props[propName])) {
+    return new Error(
+      `Invalid prop '${propName}' supplied to '${componentName}': the prop is not a valid React component`
+    );
+  } else {
+    return null;
+  }
+};
```
brianespinosa, line 16: @lordpixel  I created a custom prop-type for checking if this was a component or not. It is a little more safe than just using pt.node.

packages/cascara/src/system-components/Boundaries/Boundaries.js
```diff
@@ -1,19 +1,53 @@
...
+  // Currently ReactDOMServer does not support Suspense, but we are using it here.
+  // This is a temporary check to make sure that we check to see if we are in a browser,
+  // and if we are, then we are safe to render the boundaries. If we are not, we are on
+  // a server (like Vercel) we an render nothing because the app will render as normal.
+  // once it loads in the client browser and the app hydrates.
+  const isDOM = Boolean(
+    typeof window !== 'undefined' &&
+      window.document &&
+      window.document.createElement
   );
```
brianespinosa, line 38: Just a heads up here that we need to do this to validate if we are running in the DOM or if we are on a server. Our own apps do not need this... but our docs do. This check makes sure we do not try to render the Suspense boundary if we are on server and only try to render if we are in the browser. This is okay since we are not using SSR in our apps.

packages/cascara/src/system-components/Boundaries/Boundaries.js
lordpixel: 🙌 ok!

lordpixel (approved):

PR #254, opened 2021-09-14
chore(FDS-255): Add input date field

rociobaena:
## Input date field
The input date type is adding to our module list.
This PR includes:

- Module DataDate.
- Fixture in AllModules.fixture.js file.
- Unit test.

- [-] Component uses `React.forwardRef()` when a single DOM node is returned, _AND_ it makes React-sense to access that DOM node

### Snapshots
A new fixture for the DataDate field is created.


packages/cascara/src/modules/DataDate/index.js
```diff
@@ -0,0 +1 @@
+export { default, propTypes } from './DataDate';
```
brianespinosa, line 1: Are we actually using the prop types for this module outside of the module itself?

brianespinosa (changes requested):
@rociobaena Just a couple small things. Mainly that I think it would make sense to create a fixture file for each module which we export fixtures from to write tests against.

Your first module! 👏🏽

lordpixel (approved):

packages/cascara/src/modules/DataDate/DataDate.js
```diff
@@ -0,0 +1,76 @@
...
+    <label htmlFor={label}>
+      {label && isLabeled && <span className={styles.LabelText}>{label}</span>}
+      <Input
+        {...rest}
+        aria-label={label}
+        className={styles.Input}
+        defaultValue={value}
+        id={label}
+        max={max}
+        min={min}
```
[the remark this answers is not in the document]
rociobaena: Properties removed, and rely on ...rest instead.

packages/cascara/src/modules/DataDate/DataDate.test.js
```diff
@@ -0,0 +1,60 @@
...
+
+  describe('editing', () => {
+    // specifying ModuleSandbox will have the flag to render our input date field
+    const testId = 'default';
+
+    beforeEach(() =>
+      render(
+        <ModuleSandbox isEditing>
+          <DateDate data-testid={testId} label='Date' value={VALUE} />
+        </ModuleSandbox>
```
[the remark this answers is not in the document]
rociobaena: @brianespinosa I tried to do what you said, take a look and let me know what you think.

packages/cascara/src/modules/DataDate/index.js
rociobaena: propTypes is not being used outside of this module. Code removed.

lordpixel (approved):

packages/cascara/src/modules/DataDate/DataDate.fixture.js
```diff
@@ -0,0 +1,43 @@
...
+        <p>
+          This fixture is for testing the presentation of an input date field
+        </p>
+      </div>
+
+      <hr />
+
+      <div>
+        <h2>Display</h2>
+        <DisplayDataDate isEditing={false} value={'2021-07-24'} />
```
brianespinosa, line 31: If we have a boolean prop, we do not need to define false on it. We can safely leave off isEditing to set it as false.

packages/cascara/src/modules/DataDate/DataDate.js
```diff
@@ -0,0 +1,74 @@
...
+  value: pt.string,
+};
+
+const DataDate = ({
+  attribute,
+  isEditable = true,
+  isLabeled = true,
+  label,
+  max,
+  min,
```
brianespinosa, line 30: We removed `min` and `max` properties below on the input... but we are still destructuring them here on the DataDate function. That is effectively muting them from being passed to the Input component with `{...rest}`. We should remove them here so we do not prevent them from passing down.

packages/cascara/src/modules/DataDate/DataDate.js
```diff
@@ -0,0 +1,74 @@
...
+  /** A module can have an Attribute, which will be used as form field name */
+  attribute: pt.string,
+  /** A Module can be defined to not present an editing state */
+  isEditable: pt.bool,
+  /** Presents the input without a label. NOT USER CONFIGURABLE */
+  isLabeled: pt.bool,
+  /** A Module needs to have a unique label relative to its context */
+  label: pt.string,
+  max: pt.string,
+  min: pt.string,
```
brianespinosa, line 19: We should also remove these props from our prop-types if we are just passing them with `...rest`

packages/cascara/src/modules/DataDate/DataDate.test.js
```diff
@@ -0,0 +1,56 @@
...
+      expect(input.tagName).toMatch('SPAN');
+    });
+  });
+
+  describe('editing', () => {
+    // specifying ModuleSandbox will have the flag to render our input date field
+    const testId = 'default';
+    beforeEach(() =>
+      render(
+        <DisplayDataDate data-testid={testId} isEditing value={'2021-07-24'} />
```
brianespinosa, line 39: 🙌🏽

brianespinosa (changes requested):
@rociobaena close. A couple of those min/max props are still in our component, and we can remove the isEditing={false} in the fixture. We are ready to merge after those few changes!

brianespinosa (approved):

brianespinosa:
@rociobaena I approved but it looks like we still have a test failing. I suspect it is for a snapshot test not getting updated with the new module name.

lordpixel (dismissed):
👍

lordpixel:
@rociobaena I think we need to update snapshots

lordpixel (approved):
👍

brianespinosa (approved):

PR #255, opened 2021-09-14
chore(FDS-339): Proposal and POC for selecting rows from Table

lordpixel:
## Proposal and POC for selecting rows from Table

Resolves [FDS-339](https://espressive.atlassian.net/browse/FDS-339)

- Adds support for row selection in Table

### Dependencies

- Adds `ramda` at `v0.27.1`

### Testing

1.  `yarn cosmos`
2.  navigate to `Table -> Fixtures -> Selection` and play with the different fixtures


rociobaena:
Hi @lordpixel : 
Just a question regarding the status ALL selection
I can select all by clicking the checkbox at the top, but I can't select any of the children independently. Is this expected?
select

lordpixel:
> Hi @lordpixel :
> Just a question regarding the status ALL selection
> I can select all by clicking the checkbox at the top, but I can't select any of the children independently. Is this expected?
> select

Good catch, no! they should be intependently selectable, let me check

rociobaena (dismissed):

packages/cascara/src/ui/Table/atoms/SelectionToggle.js
```diff
@@ -2,56 +2,29 @@ import React, { useCallback, useContext } from 'react';
...
 
 import CheckBox from './Checkbox';
-import TableContext from '../context/TableContext';
+import { ModuleContext } from '../../../modules/context';
 
 const propTypes = {
   id: pt.string,
 };
 
 const SelectionToggle = ({ id }) => {
```
brianespinosa, line 11: @lordpixel I am pretty sure there is a problem with us using a module for row selection. That will effectively place the checkbox for selecting a row into the form and therefore, we now have a field that could theoretically be submitted in an API response. We will basically break functionality for submitting forms if a table has selections enabled.

Did we validate that these checkboxes are not being included when form submission takes place?

IMO it is going to be _significantly_ safer to just have a basic html checkbox that allows selection and include that at the beginning of the table row.

Maybe I am just reacting to the context being used. Let's confirm that these checkboxes do not show up when submitting forms.

packages/cascara/src/ui/Table/TableBase.js
```diff
@@ -55,9 +57,64 @@ const TableBase = ({
...
+        : [],
+    [data, uniqueIdAttribute]
+  );
+
+  /**
+   * Select row(s)
+   *
+   * @param {String|Array[String]} rowID
+   */
+  const select = useCallback(
```
brianespinosa, line 83:
```suggestion
  const rowSelect = useCallback(
```
I think it would be helpful to name these functions with more detailed descriptions than just `select`. What are we selecting? Additionally, since we should try to sort our methods alphabetically within the closure, it makes sense to try to name our functions in a way that communicates relationships between them.

The related function for this I think is the `unselect` function which could be renamed `rowUnselect` to be more clear what it does, and then they have similar name patterns for being sorted.

packages/cascara/src/ui/Table/TableBase.js
```diff
@@ -55,9 +57,64 @@ const TableBase = ({
...
+    },
+    [maxSelection, onAction, selection]
+  );
+
+  /**
+   * Unselect row(s)
+   *
+   * @param {String|Array[String]} rowID
+   */
+  const unselect = useCallback(
```
brianespinosa, line 105:
```suggestion
  const rowUnselect = useCallback(
```

brianespinosa (changes requested):
Minor function name change suggestions in the TableBase.

I think the big thing I am seeing that is likely problematic is the fact that we are using a module based form control, which might be included in form submission. We need to validate if that is happening.

brianespinosa:
Manu, I marked this PR to auto merge so once we have two approvals, and builds all pass, it will come in automatically.

packages/cascara/src/ui/Table/atoms/SelectionToggle.js
lordpixel: @brianespinosa the `checkbox` we are using for selection is NOT the `Checkbox Module` but a simple HTML element. Whilst it is part of the Form as per your suggestion, it is not included when the Form is posted. We can confirm that because the name of the checkbox is the `eID` value, and it is not present in the data posted by the form.

brianespinosa (approved):

brianespinosa:
I am setting this PR to merge as I have already approved it and it looks like Rocio did not get a chance to review PRs yesterday before getting her other PRs up.

PR #256, opened 2021-09-15
feat(FDS-344): ViewConfig P1

brianespinosa:
## ViewConfig

This PR adds our POC for `ViewConfig` with a `useViewConfig` state hook.

#### Included:
- Ability provide an agnostic list of objects for selection
- Selection of object is 100% internal
- Selected object state always returns an array of the original objects with all properties
- State hook does not expose actual state setting directly and instead exposes tightly scoped helper functions
- State can be initialized on the hook and then maintained internally

#### Not Included:
- This PR does not yet add any functionality for preserving selections to localstorage, which will need to come in a separate PR
  - Localstorage preserving may not need to happen in this component at all, keeping it more simple
- No visual support for defining presets, etc
  - Still might never be needed
  - Need to evaluate more if those things will be needed before we invest in more complexity

## JsonPlaceholder

Simplifies how the loading and empty states show, and the styles for those temp values. @rociobaena you will be interested in seeing these small changes.

 ---

### Snapshots
Updated snapshots due to changes in loading states

### Dependencies
Added `ramda` which we had not added yet and is crucial for some of the more complex functional work we need


brianespinosa:
The best way to test is the install, then run `yarn cosmos` and access the ViewConfig fixture under /private. There are multiple fixtures to check different capabilities.

lordpixel (dismissed):

PR #257, opened 2021-09-16
feat(FDS-345): [Cascara] - Replace ErrorBoundary with Boundaries

lordpixel:
## Replace ErrorBoundary with Boundaries

Resolves [FDS-345](https://espressive.atlassian.net/browse/FDS-345).

Replaces old `ErrorBoundary` component with the new `Boundaries` component.


packages/cascara/src/ui/Form/Form.js
```diff
@@ -2,7 +2,7 @@ import React, { useState } from 'react';
 import pt from 'prop-types';
 // import styles from './Form.module.scss';
 
-import Boundaries from '../../system-components/Boundaries';
```
brianespinosa, line 5: Just a heads up, the reason this was being imported directly from the file and not the index inside `/system-components` is to prevent possible future circular dependencies. If this is not causing circular dependencies now, this is okay, but we should be aware of this moving forward just in case.

On that note... it would be good to have rollup fail on circular dependencies. I will look into that. Then we can do whatever we need to do. I am creating a ticket for next month.

brianespinosa (dismissed):
This all LGTM but there are currently conflicts that need to be resolved.

Also, this needs to wait until the PR with ViewConfig comes in as it resolves an issue where I did not have the full gitignore file defined correctly for Yarn so there was a bunch of junk getting checked in to version control that we did not want, which I fixed. That should give us fewer conflicts moving forward too.

packages/cascara/src/ui/Form/Form.js
lordpixel: roger, I tested making a build and it didn't throw a warning:
image

brianespinosa: I also found that it was as easy as adding a flag to the rollup CLI so we should now be protected from circular deps in the future. 🙌🏽

lordpixel:
@brianespinosa I updated the branch 👍

brianespinosa (approved):

PR #259, opened 2021-09-16
chore(FDS-342): Add test for email input field

rociobaena:
This PR includes:
Fixture for DataEmail module
Unit test

### Snapshots
Snapshot for DataEmail test

## New Component Checklist

- [ -] Component uses `React.forwardRef()` when a single DOM node is returned, _AND_ it makes React-sense to access that DOM node


lordpixel (dismissed):
👍

packages/cascara/src/modules/DataEmail/DataEmail.js
```diff
@@ -28,13 +29,14 @@ const DataEmail = ({
   ...rest
 }) => {
   const { isEditing, formMethods } = useContext(ModuleContext);
+  const { setAriaLabel, setHtmlFor } = getHTMLLabels(isLabeled, label);
```
rociobaena, line 32: @brianespinosa I'm not sure about this name. Could you advise here?

brianespinosa: I think `getAccessibleLabelSetters`

brianespinosa (commented):
Let me know when that function helper is changed and this is ready to go!

brianespinosa (approved):
🙌🏽

PR #260, opened 2021-09-16
feature(FDS-342): Add input file field

rociobaena:
The input file type is adding to our module list.
This PR includes:
- Module DataFile
- Fixture
- Unit test

### Snapshots
A new fixture for the DataFile field is created.

## New Component Checklist

- [-] Component uses `React.forwardRef()` when a single DOM node is returned, _AND_ it makes React-sense access that DOM node


packages/cascara/src/modules/DataFile/DataFile.fixture.js
```diff
@@ -0,0 +1,37 @@
...
+  <ModuleSandbox isEditing={isEditing}>
+    <DataFile {...rest} label='Date File' />
+  </ModuleSandbox>
+);
+
+const DataFileModule = () => {
+  return (
+    <>
+      <div>
+        <h1>Date input field</h1>
```
lordpixel, line 15:
```suggestion
        <h1>File input field</h1>
```

lordpixel (approved):
@rociobaena I left a change suggestion and also approved, since it is a small change. 👍

packages/cascara/src/modules/DataFile/DataFile.fixture.js
rociobaena: Fixture updated to a new version

lordpixel (approved):

brianespinosa (approved):

PR #272, opened 2021-09-20
feat(FDS-348): [CircleCI] - Update CircleCI config to shut down workf…

lordpixel:
…low if Danger.js checks fail

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -55,14 +55,21 @@ jobs:
           key: << pipeline.parameters.installed-key >>
           paths:
             - ~/
-
+  danger:
+    <<: *job_executor
+    steps:
+      - restore_cache:
+          key: << pipeline.parameters.installed-key >>
+      - run:
+          name: Danger Checks
+          command: yarn danger ci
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -226,6 +240,9 @@ workflows:
       - install:
           <<: *install_pre_steps
           <<: *post_steps
+      - danger:
+          <<: *post_steps
+          <<: *requires_install
```
brianespinosa, line 245: @lordpixel this would need to run _before_ we do an install or any of the very long steps. That way we do not have 3 minutes of install before we find out there is a problem.

This gets tricky... because we are currently building all packages in the install step too. That might need to be separated out.

- install
- danger
- build packages

The other issue might be that we do not have danger installed as a dev dependency because we are using it in a special way inside Github Actions. There might be a circle ci orb that has danger that we could use it in a similar way.

lordpixel: roger, I will investigate (and learn) a little bit more about `Danger.js` and improve this. 👍

PR #275, opened 2021-09-21
chore(FDS-349): ViewConfig improvements

brianespinosa:
- Changes default icon for ViewConfig to an eye
- Checks for state hook before rendering or throws error
- Checks options to show loading state if undefined or empty (okay to show loading if empty as that would be invalid for this component.


brianespinosa:
Either of you could approve this. We will need to have loading support for ViewConfig, which I realized I had neglected to add. I will get us another release with this in it tonight.

rociobaena (approved):

PR #279, opened 2021-09-23
feat(FDS-351): ActionStack correct system styles

brianespinosa:
ActionStack has actual matching styles to our ViewConfig component

image

There are a number of style overlaps here, and we will definitely want to unify the styles for these and officially remove the SUI classnames because we are fighting them more than using them, but for now they are working.

> This is intentionally targeting the main branch. I think this will likely be our process for shipping anything as a patch after we have shifted back over to the app code. Ship patches direct to our main and then ship multiple `next` versions once we are in chore week.

### Snapshots

The snapshot tests that had the old vertical menu icon in them needed to be updated due to the correct 64px viewbox


lordpixel (approved):
🙌

PR #284, opened 2021-10-07
Feat/fds 343 create data io tests

lordpixel:
Resolves FDS-343.

## Add IO data type checks
Every Cascara module has its own data type, it is important for data types to remain unchanged when Form is submitted or a Table Row is edited.

This PR adds a test suite to validate module data types in two ways:

- at-rest data, data type must remain the same even for the data that isn't changed
- mutated data, data type must remain the same when data has been mutated

### Snapshots
Snapshots updated due to a change in unit tests.

> tests are passing, however, the expectations are hacked. Expect more commits to solve this.


packages/cascara/src/modules/DataSelect/DataSelect.js
```diff
@@ -51,11 +51,17 @@ const DataSelect = ({
...
       >
         {options ? (
-          options.map((option) => (
-            <option key={option.value} {...option}>
-              {option.label || option.value}
-            </option>
-          ))
+          options.map((option) =>
+            option.value === value ? (
+              <option aria-selected key={option.value} {...option} selected>
```
brianespinosa, line 56: Was eslint telling us to add `aria-selected` here? Just curious. If that is implied by the `selected` prop it would be safe to leave off. I don't think we need to change it, I just want to know if the eslint rule helped us here.

brianespinosa (changes requested):
This all looks great! I think all we probably want to do is verify if we need to have those eslint overrides. I am pretty sure we should not need them anymore with the latest eslint... but if we DO need them, those overrides should be part of the eslint config package instead so we get those for free on any kind of test files.

rociobaena (dismissed):
Manu, this is amazing :D . Good job.

PR #285, opened 2021-10-08
Chore(FDS-355) ViewConfigWarning: React does not recognize the isEditable prop on a DOM element.

rociobaena:
To get this warning in cosmos, I copied the values we get from CC and pass it as param in Cosmos.

## New Component Checklist

- [-] Includes unit tests for _ALL_ required FDS use cases
- [-] Does not mute any top level API props in React
- [-] Includes an explanation for any changes to Jest snapshots (should be a PR tag automatically added if this happens)
- [-] Component uses `React.forwardRef()` when a single DOM node is returned, _AND_ it makes React-sense to access that DOM node

### DIFF
--- a/packages/cascara/src/private/ViewConfig/ViewConfigBase.js
+++ b/packages/cascara/src/private/ViewConfig/ViewConfigBase.js
@@ -56,6 +56,8 @@ const ViewConfig = ({
   const renderOptions = (options, isSelected) =>
     options.map((option) => {
       const { currentSelection, ...setterFunctions } = state;
+      // isEditable should not propapate to MemoViewConfigItem component
+      delete option.isEditable;
       return (
         <MemoViewConfigItem
           {...option}


packages/cascara/src/private/ViewConfig/ViewConfigBase.js
```diff
@@ -56,6 +56,8 @@ const ViewConfig = ({
   const renderOptions = (options, isSelected) =>
     options.map((option) => {
       const { currentSelection, ...setterFunctions } = state;
+      // isEditable should not propapate to MemoViewConfigItem component
+      delete option.isEditable;
```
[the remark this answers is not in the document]
brianespinosa: @rociobaena I think what this is going to do is mutate the options that are being passed in to the MemoViewConfigItem. This means that the object being selected or removed will never be able to have isEditable on it. Basically we are never going to allow setting that property if we remove that value from the original object, which we do not want.

I think we may need to update the API of the ViewConfigItem to have an explicit `option` prop, instead of spreading the entire option into the component like we are doing now. That way we can always keep those props separate for use inside the ViewConfigItem.

packages/cascara/src/private/ViewConfig/ViewConfigBase.js
```diff
@@ -56,6 +56,8 @@ const ViewConfig = ({
...
       return (
         <MemoViewConfigItem
           {...option}
```
brianespinosa, line 63:
```suggestion
          option={option}
```
Essentially, the original option would become a prop instead of getting spread into the ViewConfigItem. Then we can handle those values differently inside ViewConfigItem.

brianespinosa (changes requested):
This change is having the undesired effect of removing any `isEditable` attributes from our original objects. We likely have to do a small update to the API of the internal item component instead which I made some comments on.

packages/cascara/src/private/ViewConfig/ViewConfigItem.js
```diff
@@ -16,6 +16,11 @@ const propTypes = {
   label: pt.string.isRequired,
   moveItemDown: pt.func.isRequired,
   moveItemUp: pt.func.isRequired,
+  option: pt.shape({
+    attribute: pt.string,
+    module: pt.string,
+    required: false,
+  }),
```
brianespinosa, line 23: The original option shape could be anything. We are going to cause ourselves proptype errors if we try to enforce properties inside the shape.

packages/cascara/src/private/ViewConfig/ViewConfigItem.js
```diff
@@ -77,7 +77,6 @@ const ViewConfigItem = ({
   return (
     <MenuItem
       {...state}
-      {...rest}
```
brianespinosa, line 80: I guess technically it really is this simple, huh? We can probably take out the extra memo on the object, but for now I think this is sufficient. :)

rociobaena: Brian, I tried that, but I got an error message from eslint, that's why I left it.

.changeset/plenty-tigers-marry.md
```diff
@@ -0,0 +1,5 @@
+---
+"@espressive/cascara": mejor
```
brianespinosa, line 2:
```suggestion
"@espressive/cascara": patch
```
This should be a patch change. No functionality is changing for the end user. We are only resolving a bug on our end.

rociobaena: :o Soooorry I confused the PR, changing that.

PR #286, opened 2021-10-08
chore(FDS-141) Make dangerfile not check for description comments for changes on develop/master

rociobaena:
Make danger file not check for description comments for changes on develop/master

## New Component Checklist

### DIFF
--- a/.github/workflows/danger.yml
+++ b/.github/workflows/danger.yml
@@ -2,16 +2,7 @@ name: 'Danger JS'
 on:
   pull_request:
     types:
-      [
-        assigned,
-        edited,
-        labeled,
-        opened,
-        reopened,
-        synchronize,
-        unassigned,
-        unlabeled,
-      ]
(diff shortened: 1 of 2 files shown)


.github/workflows/danger.yml
```diff
@@ -2,16 +2,7 @@ name: 'Danger JS'
...
-      [
-        assigned,
-        edited,
-        labeled,
-        opened,
-        reopened,
-        synchronize,
-        unassigned,
-        unlabeled,
-      ]
```
brianespinosa, line 14: @rociobaena is there a reason we removed the `edited` event type from this list? That will make Danger not run again if we edit the comment in our PR. This means that if a PR fails because it is missing a comment about dependencies, editing the PR comment to fix it will not trigger Danger to run again. I don't think we want that.

rociobaena:
Closed. See https://github.com/Espressive/cascara/pull/288 instead

PR #287, opened 2021-10-11
Chore(FDS-359) Move eslint-plugin-json-format to peerDependencies in the eslint config

rociobaena:
Move eslint-plugin-json-format to peerDependencies in the eslint config

### Dependencies
"eslint-plugin-json-format": "2.0.1" is included in package.json because it is a peerDependency now in eslint-config file

## New Component Checklist

### DIFF
--- a/.danger/index.ts
+++ b/.danger/index.ts
@@ -12,8 +12,6 @@ const github = {
   title: danger.github.pr.title,
 };
 
-console.log(danger);
-
 // Changed file evaluations
 const changed = {
   fixtures: modifiedFiles.filter((file) => file.includes('fixture.js')),


.changeset/silly-plums-buy.md
```diff
@@ -0,0 +1,9 @@
+---
+"docs": major
+"@espressive/cascara": major
+"@espressive/design-tokens": major
+"@espressive/icons": major
+"@espressive/eslint-config": major
+---
+
+Chore(FDS-359) Move eslint-plugin-json-format to peerDependencies in the eslint config
```
brianespinosa, line 9: @rociobaena we definitely don't want all of these packages to get a major version bump. Just the eslint config. The nice thing with Changeset files is that we can just delete this file if we do not want this to happen. We should only have the `selfish-jars-fetch.md` changeset file and remove this one.

brianespinosa (approved):
LGTM

PR #290, opened 2021-10-15
chore(FDS-141) Make danger file not check for description comments for changes on develop/master - v2

rociobaena:
Make danger file not check for description comments for changes on develop/master - v2

## New Component Checklist


.danger/index.ts
```diff
@@ -48,21 +48,26 @@ if (github.description.length < 10) {
 }
 
 // Check that someone has been assigned to this PR
-if (github.assignee === null && !isSnyk) {
+if (github.assignee === null && !isSnyk && !isCurrentBranchFromIgnored) {
```
brianespinosa, line 51: I think this was the change we needed.

I still have concerns about the logic we are using to check branches. We are only looking at the current branch and not the target. I guess this might actually work, but we need to be aware that this ALSO disables any danger checks for scenarios where `main` or `develop` go into any branch, not just each other. We do not have anything in our process at this time where that causes an issue, but I like to not have hidden gotchas in tooling that can surface later if the process updates and we have feature branches or something.

brianespinosa: I think the thing to do here is change the label of this variable so we more accurately communicate what we are doing with it. Maybe let's change `isCurrentBranchFromIgnored` to `isCurrentDevelopOrMain`. We could also probably get rid of the `ignoredBaseBranches` variable because the risk there is we have set that with a name that leads someone to believe they could add another branch there to be ignored, and we do not want to do that. Probably something like this is safer naming and sufficient:

```
const isCurrentDevelopOrMain = ['main','develop'].includes(currentBranch);
```

rociobaena: Ok @brianespinosa . Working on it.

brianespinosa (approved):
@rociobaena I like the semantics of these changes even better and that they are even more strict and explicit. This makes it very clear what these things are doing. Thank you, and congrats on your first dangerfile PR!

PR #300, opened 2021-11-05
fix(FDS-372): [ViewConfig] - allow spaces in attr search box

lordpixel:
## Resolves [FDS-372.](https://espressive.atlassian.net/browse/FDS-372)

**ViewConfig** trim() is preventing the ability to type spaces in the attribute search box.

> **Question:** A comment in DEV-17258 reads _"...putting the trim() on the search state and not the input state."_, is the search state part of Cascara?

### DIFF
--- a/packages/cascara/src/private/ViewConfig/ViewConfigBase.js
+++ b/packages/cascara/src/private/ViewConfig/ViewConfigBase.js
@@ -68,7 +68,7 @@ const ViewConfig = ({
     });
 
   const handleSearchValue = useCallback((e) => {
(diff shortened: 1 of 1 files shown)


brianespinosa (changes requested):
@lordpixel this LGTM but we do not have a changeset created for this PR so it will not cause a new build to get created.

Please create a changeset and also go ahead and mark as auto merge so when we come back to approve, this will also auto merge. 👍🏽

rociobaena (approved):

PR #302, opened 2021-11-09
feat(FDS-374): [Cascara] - Section component MVP

lordpixel:
### Section MVP

Resolves [FDS-374](https://espressive.atlassian.net/browse/FDS-374).

A new `Section` component was added according to the spec in [FDS-374](https://espressive.atlassian.net/browse/FDS-374). A `Title` component was ported from Cascara Middleware as well.

### Fixtures

New fixtures added for Section and Title

### Snapshots

New snapshots added for Section, unit tests will come when a definitive version is completed.

## New Component Checklist


packages/cascara/src/ui/Title/Title.js
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import { Boundaries } from '../../system-components';
+import { Flex } from '../../atoms';
+
+import styles from './Title.module.scss';
+import { PROP_TYPES } from './__globals';
+
+const PageTitle = ({ titlePost, titlePre, titleSub, title }) => {
+  const titleTag = title && (
+    <h3 className={styles.Title}>
```
brianespinosa, line 10: Manu, this is not something we have done in the other app where this came from... but this is a good candidate for using `Role` and `as` on the title. It means we also need to expose a new `titleAs` prop with a default of `h3`.

packages/cascara/src/ui/Section/Section.js
```diff
@@ -0,0 +1,65 @@
...
+}) => {
+  if (!ALLOWED_TAGS.includes(as)) {
+    // eslint-disable-next-line no-console -- we need this as a developer message
+    console.warn(
+      `${as} is not a valid tag, use one of: [${ALLOWED_TAGS.join(',')}]`
+    );
+  }
+
+  return (
+    <Role
```
brianespinosa, line 43: Let's also add a Boundaries component at the root of Section

.changeset/fifty-moons-build.md
```diff
@@ -0,0 +1,6 @@
+---
+"docs": patch
```
brianespinosa, line 2: @rociobaena @lordpixel there is no harm in bumping the version of the docs, as long as the version bump has the same semantics as Cascara (like this does), but note that it is not necessary as our Changeset config has Cascara and the docs linked together so they will always have the same version number at all times.

The risk here is that if we bump the semantic version of the docs higher than Cascara, it would bump Cascara too. It is bi-directional, so we should be aware of that with the changesets that the docs are linked to Cascara... and we should probably just not bump them at all and rely on that link. The changeset does not need to be changed, but this is something we should be aware of in the event we tried to bump the docs differently.

lordpixel: duly noted 👍

packages/cascara/src/ui/Section/__globals.js
```diff
@@ -0,0 +1,19 @@
...
+  isBasic: pt.bool,
+  ...TitlePT,
+};
+
+export const ALLOWED_TAGS = [
+  'section',
+  // add any other allowed tags
+];
+
+export const [DEFAULT_AS_TAG] = ALLOWED_TAGS;
```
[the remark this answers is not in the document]
lordpixel: @brianespinosa I created the `ALLOWED_TAGS ` option for us to have a way of limiting the tags to be used as `containers`. Thinking that `section`, `div`, `article`, etc. could be accepted. Now that you mention it, we probably won't need this so I removed the constant. 👍

packages/cascara/src/ui/Section/Section.module.scss
```diff
@@ -0,0 +1,12 @@
...
+  position: relative;
+  background: #fff;
+  margin: 1rem 0;
+
+  &.withPaddingAndBorder {
+    padding: 1em 1em;
+    border-radius: 0.28571429rem;
+    border: 1px solid #22242626;
+    box-shadow: 0 1px 2px 0 rgb(34 36 38 / 15%);
+  }
```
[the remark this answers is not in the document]
lordpixel: IDK why I was thinking the other way around 😄

packages/cascara/src/ui/Title/Title.js
lordpixel: @brianespinosa This is much better now, thanks for cathing this opportunity 👍

lordpixel:
> @lordpixel this PR is awesome. We are basically 95% of the way there on this... AND we are getting two new components at the same time. These are about to be used all over the place too... both in our own Cascara components, but in other work as well. I am so pumped seeing all of this code!

@brianespinosa I went through the changes you've requested, this is getting shape! 👍

packages/cascara/src/ui/Section/Section.module.scss
```diff
@@ -0,0 +1,15 @@
+.Section {
+  position: relative;
+  background: #fff;
+  margin: 1rem 0;
+  padding: 1em 1em;
```
brianespinosa, line 5: @lordpixel in the future, if we have vertical and inline values that are the same for padding (or margins) we should go ahead and just put the value once. No need to change this though. I have a feeling we are potentially going to need to tweak some of these things across Cascara in a couple months anyway so the values can come from design tokens.

Good work replicating the styes.

packages/cascara/src/ui/Section/__globals.js
```diff
@@ -0,0 +1,14 @@
+import pt from 'prop-types';
+import { PROP_TYPES as TitlePT } from '../Title/__globals';
+
+export const PROP_TYPES = {
+  /** HTML tag to allow polymorphism */
+  as: pt.string,
```
brianespinosa, line 6: @lordpixel @rociobaena technically, `as` should also allow a React component, which would not be a string. This can stay for now though as we are not doing that just yet. We still need to make a custom prop type for `as` so we can easily define it both in and outside of Cascara.

No change needed here either, just a heads up for us that we will want to replace all of these with our custom prop type when we create it. Keep defining this as `pt.string` for now.

brianespinosa (changes requested):
@lordpixel all other changes are good. Three small changes and this is good to come in. Two are things I missed in my first review:

- Changeset should be `minor` instead of `patch`

- I saw some code in Title that I missed in the first review that also needs to come out which is about Helmet, which we are not using in Cascara. Should be a simple change to just remove.

- The exports in the index.js file also need to be sorted alphabetically where we added Title.

The rest of this looks great and we can merge once those last two updates are done. Please @ me in Slack when you have a chance to do those and I can get this merged. Really looking forward to using these new components in the next release of Cascara. In fact... go ahead and create a PR from `develop` into `main` once this is merged so we can go ahead and get another minor version of Cascara out with these. If we get the Columns component PR up and merged to develop, that would also get that component included in that next minor version. 🙌🏽

rociobaena (approved):
Looks good Manu :D

brianespinosa (approved):
@lordpixel thank you!

PR #303, opened 2021-11-10
chore(FDS-375): [Role] - Switch over our components that render a DOM node at their root to use abstract components from Reakit

rociobaena:
chore(FDS-375): [Role] - Switch over our components that render a DOM node at their root to use abstract components from Reakit.
We replace div by Role, when we have a native HTML tag that has an onClick action, we use Clickable instead.

No changes in snapshots.

## New Component Checklist


rociobaena:
@brianespinosa @lordpixel could you take a special look at Pagination.js and ChatAttachment.js, please?

packages/cascara/src/structures/Admin/AdminDrawer.js
```diff
@@ -1,16 +1,21 @@
...
 const propTypes = {
   children: pt.oneOfType([pt.node, pt.array]),
+  className: pt.string,
 };
 
-const AdminDrawer = ({ children, ...rest }) => {
+const AdminDrawer = ({ children, className, ...rest }) => {
   return (
-    <div {...rest} className={styles.Drawer}>
+    <Role {...rest} className={cx(className, 'Drawer')}>
```
lordpixel, line 16: @rociobaena It's nice to see that we are swapping these, it's a much more powerful way of applying css classes 👍

packages/cascara/src/ui/Pagination/Pagination.js
```diff
@@ -169,12 +171,12 @@ const Pagination = ({ totalRecordCount, state }) => {
...
-            </div>
-          </div>
-        </div>
-      </div>
+                </Button>
+              </Role>
+            </Role>
+          </Role>
+        </Role>
+      </Role>
```
lordpixel, line 179: @rociobaena I like that we are swapping all those **divs** by **Role** components, was there any reason not to limit us to only replace the root wrapper? I think it makes sense to have them all replaced.

packages/cascara/src/ui/Pagination/Pagination.js
```diff
@@ -141,10 +143,10 @@ const Pagination = ({ totalRecordCount, state }) => {
...
                 <p className={styles.count}>{`of ${totalPages}`}</p>
               )}
-            </div>
-            <div className='field'>
-              <div className='ui small basic icon buttons'>
-                <button
+            </Role>
+            <Role className='field'>
+              <Role className='ui small basic icon buttons'>
+                <Button
```
lordpixel, line 149: I just tested this on a screen reader, Buttons need a label as they are just read as 'Button', this is out of the scope of this PR, in the future we should revisit and add the proper labels to the buttons.

packages/cascara/src/ui/Chat/ChatAttachment.js
```diff
@@ -79,7 +80,7 @@ const ChatAttachment = ({
             {type}
           </div>
           <FilesEmptyIcon outline />
-        </div>
+        </Role>
       }
       onClick={handleDownloadAttachment}
     />
```
lordpixel, line 86: Could this be something like the following?

```javascript
<Role
      as={Attachment} // please note we are using the as prop of Role
      actionable
      description={bytesToSize(size)}
      header={fileName}
      icon={
        // TODO: Extract these to CSS modules
        <Role
          style={{
            position: 'relative',
          }}
        >
          <div
            style={{
              bottom: '.375em',
              fontSize: '.75em',
              position: 'absolute',
              textAlign: 'center',
              width: '100%',
              zIndex: 1,
            }}
          >
            {type}
          </div>
          <FilesEmptyIcon outline />
        </Role>
      }
      onClick={handleDownloadAttachment}
    />
```

lordpixel (changes requested):
@rociobaena this is taking shape! 🙌  I left some comments, nothing big!

lordpixel:
@rociobaena we need to add a **changeset**, you can create one by clicking [this link](https://github.com/Espressive/cascara/new/chore/FDS-375-abstract-reackit?filename=.changeset/soft-horses-provide.md&value=---%0A%22%40espressive%2Fcascara%22%3A%20patch%0A---%0A%0Achore(FDS-375)%3A%20%5BRole%5D%20-%20Switch%20over%20our%20components%20that%20render%20a%20DOM%20node%20at%20their%20root%20to%20use%20abstract%20components%20from%20Reakit%0A) or use the link at the bottom of the message above:
image

packages/cascara/src/ui/Pagination/Pagination.js
rociobaena: Interesting, since this comes from reakit, I assume the accessibility should be fine. I'll make the change. Thank you Manu.

lordpixel: @rociobaena no change needed, I was just making a note for a future task. 👍

packages/cascara/src/ui/Chat/ChatAttachment.js
lordpixel: The above comment isn't a change request either. The code is fine 👍 

@rociobaena, @brianespinosa what do you think about the pattern described above?

rociobaena: Manu, at the beginning I thought that wouldn't work, but it is working. I think I like it.

rociobaena:
> @rociobaena we need to add a **changeset**, you can create one by clicking [this link](https://github.com/Espressive/cascara/new/chore/FDS-375-abstract-reackit?filename=.changeset/soft-horses-provide.md&value=---%0A%22%40espressive%2Fcascara%22%3A%20patch%0A---%0A%0Achore(FDS-375)%3A%20%5BRole%5D%20-%20Switch%20over%20our%20components%20that%20render%20a%20DOM%20node%20at%20their%20root%20to%20use%20abstract%20components%20from%20Reakit%0A) or use the link at the bottom of the message above: image

Manu, thank you for pointing this out. I thought we should create a new version when we finish the development week or the code should be shipped to master. Could we talk about this @lordpixel and @brianespinosa ?

packages/cascara/src/ui/Pagination/Pagination.js
rociobaena: Agree, we could create a task to check all the proper labels not only to buttons.

packages/cascara/src/ui/Pagination/Pagination.js
rociobaena: Manu, I have doubts about this change too, that's why I added the comment to put special attention to this component. I added the Role to all these components because they are containers at the end. I know that piece of code is not visible enough, but when I saw this, it is a big container containing two more components, that's why I decided to use Role here too. 
image

packages/cascara/src/ui/Chat/ChatAttachment.js
lordpixel: awesome, if it works and everybody agrees, let's keep it 🙌

packages/cascara/src/atoms/Flex/Flex.js
```diff
@@ -54,7 +55,7 @@ const Flex = ({
     ...getClassOptions(vAlignOptions, vAlign, 'vAlign'),
   });
 
-  return <div className={classList}>{children}</div>;
+  return <Role className={classList}>{children}</Role>;
```
brianespinosa, line 58:
```suggestion
  return <Role {...rest} as={as} className={classList} >{children}</Role>;
```
We need to expose the `as` prop here, and spread the rest of the props. The rest spread is very important for component compatibility with `as`

packages/cascara/src/atoms/Flex/FlexItem.js
```diff
@@ -35,7 +36,7 @@ const FlexItem = ({
   // Might need to expand this for other than strings later
   const isValidElement = typeof children !== 'string';
 
-  const element = isValidElement ? children : <div>{children}</div>;
+  const element = isValidElement ? children : <Role>{children}</Role>;
```
brianespinosa, line 39: @rociobaena this is a unique scenario where we do not need `Role`. We are only rendering a `div` in the scenario where we have a string by itself and we need to add other flex classes to it. If there is a dom element as the child, then we use that component itself. So the div is only being used for wrapping strings, in which case, we do not need any of the benefits of Role for using `as`.

packages/cascara/src/modules/DataCheckbox/DataCheckbox.js
```diff
@@ -61,9 +62,9 @@ const DataCheckbox = ({
   // Do not render an editable input if the module is not editable
   return (
     <ModuleErrorBoundary>
-      <div className={styles.Checkbox}>
+      <Role className={styles.Checkbox}>
```
brianespinosa, line 65: @rociobaena everything in `cascara/src/modules` should _not_ be using Role. These are not exported, root level elements. All of the modules should be returned back to the `div`. Remember... this is only for exported, root-level components so we can use `as` on them.

packages/cascara/src/private/Section/Section.js
```diff
@@ -13,7 +14,7 @@ const Section = ({ children, footer, header }) => {
   return (
     <section className={styles._}>
       {header && <header className={styles.Header}>{header}</header>}
-      <div className={styles.Content}>{children}</div>
+      <Role className={styles.Content}>{children}</Role>
```
brianespinosa, line 17: @rociobaena this component is not the root. It can stay as `div`

The component in Section that needs to be Role is actually the one on line 15. 
```
<Role as={as} /> // default prop for as = 'section'

brianespinosa (changes requested):
@rociobaena there are a lot of changes that go too far here. We need to get those rolled back first. Only exported components should get this update, not every component, and only the very root node that renders a DOM element should get a Role component, not every div inside of the component. We need to expose the `as` prop on all of these when we add Role.

It might be easier to go clean away anything that is not a root export first so we are looking at a more limited list of changes. Then we can start looking at the smaller details like merging props for `classname`, `style`, and exposing `as` with a default prop... while also doing a rest spread.

packages/cascara/src/atoms/Flex/FlexItem.js
rociobaena: Brian, I think I got it (for the third time :P ) Also, In this case, it is not need because we should have the Role in the **Flex** component, that will create our root component

rociobaena:
There were a lot changes that we don't need. A new open PR with the requested changes are here instead: https://github.com/Espressive/cascara/pull/307

PR #305, opened 2021-11-11
feat(DEV-373):  [Cascara] - Columns component mvp

lordpixel:
## Resolves [FDS-373](https://espressive.atlassian.net/browse/FDS-373).

Add a new component called Columns that when wrapped around children will cause those children to display in columns using CSS.
> this is NOT grid. 

### Snapshots
Snapshots created for the new component.

### Fixtures
Fixtures created for the new component.

## New Component [Checklist](url)

### DIFF
--- a/.changeset/short-rabbits-joke.md
+++ b/.changeset/short-rabbits-joke.md
@@ -1,0 +1,6 @@
+---
+"docs": patch
+"@espressive/cascara": patch
+---
+
+feat(DEV-373):  [Cascara] - Columns component mvp
--- a/.changeset/short-rabbits-joke.md
+++ b/.changeset/short-rabbits-joke.md
@@ -1,6 +1,6 @@
 ---
-"docs": patch
-"@espressive/cascara": patch
+'docs': patch
+'@espressive/cascara': patch
 ---
 
-feat(DEV-373):  [Cascara] - Columns component mvp
+feat(FDS-373): [Columns] - Component mvp
--- a/packages/cascara/src/ui/Columns/Columns.js
+++ b/packages/cascara/src/ui/Columns/Columns.js
@@ -9,26 +9,45 @@ import styles from './Columns.module.scss';
 const cx = classNames.bind(styles);
 
 const propTypes = {
+  /** HTML tag to allow polymorphism */
+  as: pt.string,
   /** The section content */
   children: pt.oneOfType([pt.arrayOf(pt.node), pt.node]),
   /** Columns can have css class name */
   className: pt.string,
   /** The number of columns to allocate */
   count: pt.number,
+  /** Columns can have inline styles */
+  style: pt.shape(pt.object),
 };
 
-const Columns = ({ count = 2, children, className, ...rest }) => (
-  <Boundaries>
-    <Role
-      className={cx('Columns', className)}
-      style={{ columnCount: count }}
-      {...rest}
-    >
-      {children}
-    </Role>
-  </Boundaries>
-);
+const Columns = ({
+  as = 'div',
+  count = 2,
+  children,
+  className,
+  style,
+  ...rest
+}) => {
+  let mergedStyle = {
+    columnCount: count,
+  };
 
+  if (style) {
+    mergedStyle = {
+      ...style,
+      ...mergedStyle,
+    };
+  }
+
+  return (
+    <Boundaries>
+      <Role className={cx('Columns', className)} style={mergedStyle} {...rest}>
+        {children}
+      </Role>
+    </Boundaries>
+  );
+};
 Columns.propTypes = propTypes;
 
 export { propTypes };
--- a/packages/cascara/src/ui/Columns/Columns.js
+++ b/packages/cascara/src/ui/Columns/Columns.js
@@ -16,14 +16,14 @@ const propTypes = {
   /** Columns can have css class name */
   className: pt.string,
   /** The number of columns to allocate */
-  count: pt.number,
+  count: pt.oneOfType([pt.number, pt.oneOf(['auto'])]),
   /** Columns can have inline styles */
   style: pt.shape(pt.object),
 };
 
 const Columns = ({
   as = 'div',
-  count = 2,
+  count = 'auto',
   children,
   className,
   style,
@@ -33,7 +33,7 @@ const Columns = ({
     columnCount: count,
   };
 
-  if (style) {
+  if (typeof count === 'number' && style) {
     mergedStyle = {
       ...style,
       ...mergedStyle,
--- a/packages/cascara/src/ui/Columns/Columns.js
+++ b/packages/cascara/src/ui/Columns/Columns.js
@@ -1,0 +1,35 @@
+import React from 'react';
+import pt from 'prop-types';
+import { Role } from 'reakit';
+import classNames from 'classnames/bind';
+
+import Boundaries from '../../system-components/Boundaries';
+import styles from './Columns.module.scss';
+
+const cx = classNames.bind(styles);
+
+const propTypes = {
+  /** The section content */
+  children: pt.oneOfType([pt.arrayOf(pt.node), pt.node]),
+  /** Columns can have css class name */
+  className: pt.string,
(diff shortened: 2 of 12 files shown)


packages/cascara/src/ui/Columns/Columns.module.scss
```diff
@@ -0,0 +1,15 @@
+$masonry-gutter: 2em;
+$widget-border-radius: 0.5em;
+$widget-background: #f0f0f0;
+
+.Columns {
```
lordpixel, line 5: @brianespinosa Let me know if this is what we are looking for, here's a screenshot from a standalone window in Cosmos for this change:
image

packages/cascara/src/ui/Columns/Columns.js
```diff
@@ -0,0 +1,35 @@
...
+  children: pt.oneOfType([pt.arrayOf(pt.node), pt.node]),
+  /** Columns can have css class name */
+  className: pt.string,
+  /** The number of columns to allocate */
+  count: pt.number,
+};
+
+const Columns = ({ count = 2, children, className, ...rest }) => (
+  <Boundaries>
+    <Role
```
brianespinosa, line 22: @lordpixel because we are using the Role component, I think we should be exposing the `as` prop (with a default as well) to developers. The default `as` in this scenario should be `'div'`
```suggestion
    <Role
      as={as}
```

packages/cascara/src/ui/Columns/Columns.js
```diff
@@ -0,0 +1,35 @@
...
+      className={cx('Columns', className)}
+      style={{ columnCount: count }}
```
brianespinosa, line 24: Manu, for now let's not change this... but we may need to consider the possibility that we allow direct, inline styles in some cases if someone needs to fix something. In that scenario, we might need to have a solution to merge objects if a user defines a `style` prop. In the current scenario, technically that will not work and we will throw any of their styles away.

That can be discussed separately, but we should be aware of this. We might be giving a weird experience to users as some places they can use `style` where we do a rest spread, and others, like this, they cannot.

.changeset/short-rabbits-joke.md
```diff
@@ -0,0 +1,6 @@
+---
+"docs": patch
+"@espressive/cascara": patch
+---
+
+feat(DEV-373):  [Cascara] - Columns component mvp
```
brianespinosa, line 6: @lordpixel I think you did the component that was being added before in brackets, right? That seems better to me so we can see all the different component related things together in the release notes.

I am also creating a separate task before I forget that we need to get the latest release changelog copied to the package README each time so when you look at the releases in Github you can see everything that was included in that release. That is not part of this ticket... your changeset just reminded me.

packages/cascara/src/ui/Columns/Columns.mdx
```diff
@@ -0,0 +1,16 @@
+---
+title: Columns
+propTable: Columns.js
+---
+
+```jsx
+---
+title: An example use of Columns with default props (2 columns)
```
brianespinosa, line 8:
```suggestion
title: An example use of Columns with automatic columns based on width, which will change depending on the width of the Columns container.
```

packages/cascara/src/ui/Columns/Columns.module.scss
```diff
@@ -0,0 +1,15 @@
+$masonry-gutter: 2em;
```
brianespinosa, line 1:
```suggestion
$masonry-gutter: 2rem;
```

packages/cascara/src/ui/Columns/Columns.module.scss
```diff
@@ -0,0 +1,15 @@
...
+$widget-border-radius: 0.5em;
+$widget-background: #f0f0f0;
```
brianespinosa, line 3:
```suggestion
```

brianespinosa (changes requested):
@lordpixel small changes but otherwise this basically is ready to go too. Great work! 🙌🏽

packages/cascara/src/ui/Columns/Columns.module.scss
brianespinosa: Yes. The column sizing looks great. I am marking this as resolved so it is out of the way when you take a look at the rest of the comments. 👍🏽

packages/cascara/src/ui/Columns/Columns.js
brianespinosa: Let's not make a change here, but we are going to need a style merge helper function to make this easier to do.

packages/cascara/src/ui/Columns/Columns.module.scss
brianespinosa: These are not being used

.changeset/short-rabbits-joke.md
lordpixel: You are right, I used this Cascara scope because Cascara was the package the Columns component was added to. I agree with you that this should be the component name, which makes me think: Could we use some kind of namespace? 

i.e. feat(FDS-373): [Cascara - Columns] -  Columns component mvp

The idea is to have the name of the package like `Cascara`, `Tokens`, `Icons` etc and then the name of the piece of code that was changed. Let me update the changeset 👍

packages/cascara/src/ui/Columns/Columns.js
lordpixel: @brianespinosa I did not commit the suggestion because we also needed to add props definition for `as`, so I went and did it locally. 👍

packages/cascara/src/ui/Columns/Columns.js
lordpixel: @brianespinosa I did not create the helper function, as you said, we need to discuss it separately. But the component now accepts an extra prop `style` that is merged with the internal inline style. 👍

packages/cascara/src/ui/Columns/Columns.js
```diff
@@ -0,0 +1,54 @@
...
+
+const propTypes = {
+  /** HTML tag to allow polymorphism */
+  as: pt.string,
+  /** The section content */
+  children: pt.oneOfType([pt.arrayOf(pt.node), pt.node]),
+  /** Columns can have css class name */
+  className: pt.string,
+  /** The number of columns to allocate */
+  count: pt.number,
```
brianespinosa, line 19: I think I missed this on the last review. Count should be either a number or the string 'auto'

packages/cascara/src/ui/Columns/Columns.js
```diff
@@ -0,0 +1,54 @@
...
+  /** Columns can have inline styles */
+  style: pt.shape(pt.object),
+};
+
+const Columns = ({
+  as = 'div',
+  count = 2,
```
brianespinosa, line 26:
```suggestion
  count = 'auto',
```
The default prop for columns should be 'auto' not 2

packages/cascara/src/ui/Columns/Columns.js
```diff
@@ -0,0 +1,54 @@
...
+  let mergedStyle = {
+    columnCount: count,
+  };
+
+  if (style) {
+    mergedStyle = {
+      ...style,
+      ...mergedStyle,
+    };
+  }
```
brianespinosa, line 41: @lordpixel there needs to be some sort of prop check for columnCount. If `typeOf columnCount === number` then we merge the styles. Otherwise we just pass the styles directly. We are not supporting defining column width at this time, only using the default column width (auto) or a count.

brianespinosa (changes requested):
A couple small changes needed in the column setting logic, and the default prop and prop types so that 'auto' is allowed as the exact string. All other changes look good and should be ready to come in!

packages/cascara/src/ui/Columns/Columns.js
lordpixel: ok, changes applied 👍

brianespinosa (approved):
@lordpixel this looks great. It also solves the edge cases so we don't have to come back to them later.

PR #306, opened 2021-11-11
feat(FDS-371): [Cascara] - improve fallback component

lordpixel:
### Resolves [FDS-371](https://espressive.atlassian.net/browse/FDS-371).

Improve default Boundaries FallbackComponent to only show the error message in the UI during development environments with support for showing them with a localstorage key.

### Fixtures
Fixture for Boundaries updated with an error message.

## New Component Checklist

### DIFF
--- a/packages/cascara/src/system-components/Boundaries/fallbacks/ErrorFallback.js
+++ b/packages/cascara/src/system-components/Boundaries/fallbacks/ErrorFallback.js
@@ -1,6 +1,7 @@
-import React, { useCallback, useState } from 'react';
-import { Checkbox, Role } from 'reakit';
+import React, { useCallback } from 'react';
 import pt from 'prop-types';
+
+import Button from '../../../atoms/Button/Button';
 
 const propTypes = {
   // eslint-disable-next-line react/forbid-prop-types -- error objects
@@ -9,44 +10,25 @@ const propTypes = {
 };
 
 const ErrorFallback = ({ error, resetErrorBoundary }) => {
-  const checkedFromLS =
-    window.localStorage.getItem('ESPRESSIVE_SHOW_ERRORS') === 'true';
-  const [checked, setChecked] = useState(checkedFromLS);
-  const showError = process.env.NODE_ENV === 'development' || checked;
-
-  const hangleCheckboxClick = useCallback(() => {
-    setChecked(!checked);
-  }, [checked]);
-
-  const hangleTryAgainClick = useCallback(() => {
-    window.localStorage.setItem('ESPRESSIVE_SHOW_ERRORS', checked);
-
+  const showError = process.env.NODE_ENV === 'development';
+  const handleTryAgain = useCallback(() => {
     if (resetErrorBoundary) {
       resetErrorBoundary();
     }
-  }, [checked, resetErrorBoundary]);
+  }, [resetErrorBoundary]);
 
   return (
-    <Role className='ui tiny error message' role='alert'>
+    <div className='ui tiny error message' role='alert'>
       <div className='header'>Something went wrong:</div>
-      {showError && <pre>{error?.message}</pre>}
-      <br />
-      <button
-        className='ui negative button'
-        onClick={hangleTryAgainClick}
-        type='button'
-      >
+      {showError && (
+        <p>
+          <pre>{error?.message}</pre>
+        </p>
+      )}
+      <Button onClick={handleTryAgain} outcome='negative' type='button'>
         Try again
-      </button>
-      <label className='ui' htmlFor='chkAlwaysShowErrors'>
-        <Checkbox
-          checked={checked}
-          id='chkAlwaysShowErrors'
-          onChange={hangleCheckboxClick}
-        />{' '}
-        Always show errors
-      </label>
-    </Role>
+      </Button>
+    </div>
   );
 };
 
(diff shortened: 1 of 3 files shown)


packages/cascara/src/system-components/Boundaries/fallbacks/ErrorFallback.js
```diff
@@ -8,18 +9,44 @@ const propTypes = {
...
+  }, [checked, resetErrorBoundary]);
+
   return (
-    <div className='ui tiny error message' role='alert'>
+    <Role className='ui tiny error message' role='alert'>
       <div className='header'>Something went wrong:</div>
-      <pre>{error?.message}</pre>
+      {showError && <pre>{error?.message}</pre>}
+      <br />
       <button
```
rociobaena, line 34: @lordpixel should be using Button from reakit instead?

packages/cascara/src/system-components/Boundaries/fallbacks/ErrorFallback.js
```diff
@@ -8,18 +9,44 @@ const propTypes = {
...
         className='ui negative button'
-        onClick={resetErrorBoundary}
+        onClick={hangleTryAgainClick}
```
rociobaena, line 36: Typo: han**g**leTryAgainClick :D

packages/cascara/src/system-components/Boundaries/fallbacks/ErrorFallback.js
```diff
@@ -8,18 +9,44 @@ const propTypes = {
...
+    }
+  }, [checked, resetErrorBoundary]);
+
   return (
-    <div className='ui tiny error message' role='alert'>
+    <Role className='ui tiny error message' role='alert'>
       <div className='header'>Something went wrong:</div>
-      <pre>{error?.message}</pre>
+      {showError && <pre>{error?.message}</pre>}
+      <br />
```
brianespinosa, line 33:
```suggestion
      {showError && <p><pre>{error?.message}</pre></p>}
```

packages/cascara/src/system-components/Boundaries/fallbacks/ErrorFallback.js
brianespinosa: @lordpixel Yes, that would be good, but I think we could also just swap this over officially for the Button component from Cascara instead... which is using Reakit. We could also remove the classname values and use the props on that Button instead. If the negative type styles are not there yet, let's still use that type and fix the styles as a separate task.

packages/cascara/src/system-components/Boundaries/fallbacks/ErrorFallback.js
brianespinosa:
```suggestion
        onClick={handleTryAgain}
```

packages/cascara/src/system-components/Boundaries/fallbacks/ErrorFallback.js
```diff
@@ -8,18 +9,44 @@ const propTypes = {
...
       </button>
-    </div>
+      <label className='ui' htmlFor='chkAlwaysShowErrors'>
+        <Checkbox
+          checked={checked}
+          id='chkAlwaysShowErrors'
+          onChange={hangleCheckboxClick}
+        />{' '}
+        Always show errors
+      </label>
```
brianespinosa, line 48: @lordpixel what use case do we feel we are solving with this checkbox? We do not want end users to see this to be able to opt in to messages. That will basically make this checkbox show for everyone instead of the error message in production, which we don't want. We don't want customers opting in to these types of errors on screen. At least not by default. If we show this to everyone, I think we will just have everyone opting in to these everywhere.

I like the addition to have a localstorage value to enable if we needed to look at something in production, but having UI to do that I don't think is the goal here.

Since this checkbox was not in scope for the ticket description, I want to make sure I am not missing anything here. Maybe you have a good use case that you think we should actually have this.

brianespinosa (changes requested):

packages/cascara/src/system-components/Boundaries/fallbacks/ErrorFallback.js
lordpixel: @brianespinosa The checkbox is there because that is what I understood from the task title:
`Improve default Boundaries FallbackComponent to only show the error message in the UI during development environments with support for showing them with a localstorage key`.

The approach I decided to go with is a checkbox that controls the localStorage value. I apologize if that is not what we want. Let me remove the checkbox and the localStorage key 👍

packages/cascara/src/system-components/Boundaries/fallbacks/ErrorFallback.js
```diff
@@ -8,18 +9,44 @@ const propTypes = {
...
+    window.localStorage.setItem('ESPRESSIVE_SHOW_ERRORS', checked);
+
+    if (resetErrorBoundary) {
+      resetErrorBoundary();
+    }
+  }, [checked, resetErrorBoundary]);
+
   return (
-    <div className='ui tiny error message' role='alert'>
+    <Role className='ui tiny error message' role='alert'>
```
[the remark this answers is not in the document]
lordpixel: I did not commit the suggestion because I had more changes in my local, but this change has been applied. 👍

packages/cascara/src/system-components/Boundaries/fallbacks/ErrorFallback.js
lordpixel: I did not commit the suggestion because I had more changes in my local, but this change has been applied. 👍

packages/cascara/src/system-components/Boundaries/fallbacks/ErrorFallback.js
lordpixel: oh my typos!

brianespinosa (approved):
@lordpixel LTM!

PR #308, opened 2021-11-17
doc(FDS-377): [Docs] use Cascara's AdminStructure in Docs

lordpixel:
Resolves FDS-377.
image

Use AdminStructure in Docs. Remove old Admin.


docs/src/pages/_app.js
```diff
@@ -35,52 +34,93 @@ const MyApp = ({ Component, pageProps }) => {
...
-                  key={router.query.mdx + componentProps}
+      }
+      nav={<Nav {...pageProps} />}
+      theme={theme}
+    >
+      <AdminStructure.Main>
+        <Component {...pageProps} />
+      </AdminStructure.Main>
+      <AdminStructure.Drawer>
+        <div style={{ padding: '2rem' }}>
```
brianespinosa, line 59: @lordpixel this padding I do not think is needed in our new Drawer, so it might also be able to go away.

I think this is looking good so far.

That aside... I think we might want to remove the motion stuff for now with the animations. I think that might be something we include inside of our structure components themselves at a later date.

Good work so far!

lordpixel: Nice, these changes have been apploied, thank you @brianespinosa

lordpixel:
@brianespinosa we are seeing the following error message when running docs:

```
ReferenceError: window is not defined
    at Object.useMediaQuery (webpack-internal:///../packages/cascara/dist/cjs/DataUrl-343de13f.js:223:20)
    at AdminProvider (webpack-internal:///../packages/cascara/dist/cjs/index.js:20:421)
    at processChild (/Users/manuramirez/code/cascara/node_modules/react-dom/cjs/react-dom-server.node.development.js:3043:14)
    at resolve (/Users/manuramirez/code/cascara/node_modules/react-dom/cjs/react-dom-server.node.development.js:2960:5)
    at ReactDOMServerRenderer.render (/Users/manuramirez/code/cascara/node_modules/react-dom/cjs/react-dom-server.node.development.js:3435:22)
    at ReactDOMServerRenderer.read (/Users/manuramirez/code/cascara/node_modules/react-dom/cjs/react-dom-server.node.development.js:3373:29)
    at renderToString (/Users/manuramirez/code/cascara/node_modules/react-dom/cjs/react-dom-server.node.development.js:3988:27)
    at Object.renderPage (/Users/manuramirez/code/cascara/node_modules/next/dist/next-server/server/render.js:54:851)
    at Function.getInitialProps (webpack-internal:///../node_modules/next/dist/pages/_document.js:135:19)
    at Function.getInitialProps (webpack-internal:///./src/pages/_document.js:20:85)
    at loadGetInitialProps (/Users/manuramirez/code/cascara/node_modules/next/dist/next-server/lib/utils.js:5:101)
    at renderToHTML (/Users/manuramirez/code/cascara/node_modules/next/dist/next-server/server/render.js:54:1142)
    at async /Users/manuramirez/code/cascara/node_modules/next/dist/next-server/server/next-server.js:107:97
    at async __wrapper (/Users/manuramirez/code/cascara/node_modules/next/dist/lib/coalesced-function.js:1:330)
    at async DevServer.renderToHTMLWithComponents (/Users/manuramirez/code/cascara/node_modules/next/dist/next-server/server/next-server.js:132:387)
    at async DevServer.renderToHTML (/Users/manuramirez/code/cascara/node_modules/next/dist/next-server/server/next-server.js:133:522)
```

Any idea why we are seeing this error? it seems to be caused by the `useMediaQuery` hook.

lordpixel:
> 

I think it's because we are using SSR, and window is not present.
image

brianespinosa:
> I think it's because we are using SSR, and window is not present.

Correct. The solution for these scenarios is to check if window is present before we try to access window. In the case of this hook, it might be as simple as `window?.matchedMedia` but that depends on how the rest of the code in that hook reacts to undefined. I think this would be the best approach or using the nullish coalescing operator to define a default where the hook will always work even if no window is present.

lordpixel:
> > I think it's because we are using SSR, and window is not present.
> 
> Correct. The solution for these scenarios is to check if window is present before we try to access window. In the case of this hook, it might be as simple as `window?.matchedMedia` but that depends on how the rest of the code in that hook reacts to undefined. I think this would be the best approach or using the nullish coalescing operator to define a default where the hook will always work even if no window is present.

@brianespinosa this was solved by moving some of the code that attempts to access the `window` object into a `useEffect` call.

packages/cascara/src/hooks/useMediaQuery.js
```diff
@@ -3,21 +3,28 @@
...
+
   // We need to set state here so that the hook will return the new value
-  const [match, setMatch] = useState(Boolean(mediaQuery.matches));
+  const [match, setMatch] = useState();
+
+  useEffect(() => {
+    const mediaQuery = window.matchMedia(query);
+
+    setMediaQuery(mediaQuery);
+  }, [query]);
```
brianespinosa, line 15: @lordpixel is there a reason not to just move the logic from this useEffect inside of the useEffect that is directly below on line 17? It all needs to be done at the same time, and they will not be done separately. If this needs the useEffect, I think it makes sense to have it all inside the same effect hook.

lordpixel: @brianespinosa thanks for your help on this one. 👍

lordpixel:
@brianespinosa as far as I remember, Admin is only used in Docs and test apps, hence we are only removing the component forlder and its exports/imports Instead of deprecating the component. Is that ok?

brianespinosa:
@lordpixel there are some conflicts on AdminDrawer here that need to get resolved. Otherwise I think this generally looks correct. I will give this a detailed test early tomorrow morning before our standup. Once that conflict is resolved, this can probably come in if I don't run into any issues in testing. 👍🏽 

Thank you for doing this. I can see the docs for the structure component and slot components can be improved based on seeing your experience converting this, so that is an area we can invest further in other tickets.

brianespinosa (changes requested):
@lordpixel from a code perspective, I think this is all fine. I am noticing though that we are now missing the props table on all of our components. This is a pretty big thing to be missing. We should address that before we allow this to come in.

Also... I know there is a file version of our Cascara logo somewhere, as we are using that file on the README at the repo root. We should find that and get it imported for the header logo so that is also working.

Third, there are strange styles going on for the Nav items. We can remove the old styles that were being used for those. I think we can just remove the CSS Modules being used on docs/src/components/Nav/NavItem

Other than those two things... I THINK this is okay. There are some strange styles going on in the nav items,

lordpixel:
> @lordpixel from a code perspective, I think this is all fine. I am noticing though that we are now missing the props table on all of our components. This is a pretty big thing to be missing. We should address that before we allow this to come in.
> 
> Also... I know there is a file version of our Cascara logo somewhere, as we are using that file on the README at the repo root. We should find that and get it imported for the header logo so that is also working.
> 
> Third, there are strange styles going on for the Nav items. We can remove the old styles that were being used for those. I think we can just remove the CSS Modules being used on docs/src/components/Nav/NavItem
> 
> Other than those three things... I THINK this is okay. 👍🏽

@brianespinosa Thanks for catching these. I am working on the fixes. 

Regarding the styles: If we remove the old style we get this:
image

One thing I noticed is the old `NavList` is using a ul tag, whereas the new one is using Role from reakit, that's why they seem to break the styling.

rociobaena (approved):
This is awesome 😎

.changeset/calm-eyes-return.md
```diff
@@ -0,0 +1,6 @@
+---
+"docs": patch
+"@espressive/cascara": patch
+---
+
+doc(FDS-377): [Docs] use Cascara's AdminStructure in Docs
```
brianespinosa, line 6: @lordpixel @rociobaena just a heads up... we do not want to include changesets for Docs. The reason for this is that Docs and Cascara are linked in our Changeset config. This means that when the Cascara version changes, it syncs the version of Docs to keep in line with it. If we do a version bump of the docs, we will actually also bump the version of Cascara inadvertently.

We should consider a docs update similar to any other internal tooling, which we do not want to version bump as those versions are not for us internally, but instead they are used for anyone external to our project to know what we have changed.

app-tests/next-layout-admin/src/pages/index.js
```diff
@@ -2,24 +2,24 @@ import '@espressive/legacy-css';
...
 const Home = () => {
   return (
-    <Admin
+    <AdminStructure
       header={
-        <Admin.Header>
+        <AdminStructure.Header>
           <h1>{pkg.name}</h1>
-        </Admin.Header>
+        </AdminStructure.Header>
```
brianespinosa, line 13: If we are updating this example, we should also make sure that it follows the correct APIs as AdminStructure slot components. The h1 here as a child is not needed. Instead `pkg.name` can be passed as a prop to `AdminStructure.Header` not as a child. And the `h1` tag is not needed at all.

docs/public/cascara_meta.svg
```diff
@@ -1 +1 @@
...
...43"/><stop offset="1" stop-color="#c51653"/></linearGradient><linearGradient id="C" x1="83.742%" x2="17.104%" y1="108.444%" y2="-6.98%"><stop offset="0" stop-color="#00873e"/><stop offset="1" stop-color="#006129"/></linearGradient><linearGradient id="D" x1="50%" x2="50%" y1="100%" y2="0%"><stop offset="0" stop-color="#b4104b"/><stop offset="1" stop-color="#d91c5c"/></linearGradient><g transform="translate(60 30)" fill-rule="evenodd"><path d="M15 12.5a7.5 7.5 0 1 0 15 0 7.5 7.5 0 1 0-15 0" fill="url(#A)"/><path d="M20 25a10 10 0 1 0 20 0 10 10 0 1 0-20 0" fill="url(#B)"/><path d="M2.495 0C1.718.006.942.062.171.171c-.733 5.225.892 10.721 4.914 14.743s9.519 5.649 14.744 4.915c.733-5.225-.892-10.722-4.914-14.744C11.547 1.718 7.147.031 2.737 0z" fill="url(#C)"/><path d="M0 27.5C0 34.403 5.597 40 12.5 40S25 34.403 25 27.5 19.403 15 12.5 15 0 20.597 0 27.5" fill="url(#D)"/><text fill="#333" font-family="AgletSlab-Semibold, Aglet Slab" font-size="42" font-weight="500" letter-spacing="-1.768"><tspan x="44.1" y="34.5">C</tspan><tspan font-size="38" letter-spacing="-1.6" x="68.918" y="34.5">a</tspan><tspan font-size="38" letter-spacing="-1.2" x="88.522" y="34.5">scara</tspan></text></g></svg>
```
brianespinosa, line 1: This SVG file is used in the README.md file. We should not modify it as this will break how the README looks with our logo on dark screens, etc. We should instead make a new, separate SVG file based on the SVG info inside the Logo.js file that was previously being used.

It probably also makes sense to move this file that it used specifically for the README and put it in the repo root with the README. It is not very obvious that the README is using this file that is located within a package.

app-tests/cra-layout-admin/src/App.js
```diff
@@ -1,16 +1,16 @@
...
 const App = () => {
   return (
-    <Admin
+    <AdminStructure
       header={
-        <Admin.Header>
+        <AdminStructure.Header>
           <h1>{process.env.REACT_APP_NAME}</h1>
-        </Admin.Header>
+        </AdminStructure.Header>
```
brianespinosa, line 10: This API change should match the API change being made in the next-layout-admin template for the header.

docs/src/components/Nav/NavLinkNext.js
```diff
@@ -0,0 +1,23 @@
+import pt from 'prop-types';
+import Link from 'next/link';
+
+const propTypes = {
+  children: pt.oneOfType([pt.arrayOf(pt.node), pt.node]),
+  isActive: pt.bool,
+};
+
+const NavLinkNext = ({ children, isActive, ...rest }) => {
```
brianespinosa, line 9:
```suggestion
// This component is needed as a compatibility layer inside of AdminStructure.Nav due to the odd way that
// Link works inside of `next/link` where it does not render an element itself. This component allows us to 
// maintain a structure that is consistent with other link components, and direct props into the expected location.
const NavLinkNext = ({ children, isActive, ...rest }) => {
```

brianespinosa: I think this note will be helpful for our future selves if someone comes along and sees this component later. I'd recommend just committing this suggestion directly to make things easier.

packages/cascara/src/hooks/useMediaQuery.js
```diff
@@ -3,22 +3,33 @@
 import { useEffect, useState } from 'react';
 
 const useMediaQuery = (query, ifTrueFunc, ifFalseFunc) => {
-  const mediaQuery = window.matchMedia(query);
+  const isDOM = Boolean(
+    typeof window !== 'undefined' &&
+      window.document &&
+      window.document.createElement
+  );
```
brianespinosa, line 10: Could we find the other place we were using this isDOM function and place it in a location where we can reuse it? I know there are going to be other scenarios where we need it moving forward. We might as well make sure we are reducing boilerplate now.

If any of those other places exist inside of the Docs project, we can disregard them for now. But anywhere we are using this in Cascara, we should just have one helper function.

brianespinosa (changes requested):
@lordpixel mostly ready to go. There are still a number of very small changes we need to make. The majority of this looks basically ready now. Most of this I think is last little cleanup and we can merge.

Again... very excited to get this one in!

.changeset/calm-eyes-return.md
lordpixel: Makes sense, removed, thanks for the heads up 👍

docs/public/cascara_meta.svg
lordpixel: SVG file restored and moved to the root, we also have a new SVG logo in Public which is based on the existing Logo component.

packages/cascara/src/hooks/useMediaQuery.js
lordpixel: There was only one copy of the function in Boundaries. Function moved to /lib folder and both components updated.

lordpixel:
> @lordpixel mostly ready to go. There are still a number of very small changes we need to make. The majority of this looks basically ready now. Most of this I think is last little cleanup and we can merge.
> 
> Again... very excited to get this one in!

@brianespinosa thanks for your time reviewing this PR, Docs have a much better shape! This is ready for re-review.

brianespinosa (approved):
Thanks @lordpixel 🙌🏽

PR #310, opened 2021-11-22
chore(FDS-376): [Cascara] - Some official custom prop types

rociobaena:
Export custom prop types

## New Component Checklist


lordpixel (changes requested):
Hi @rociobaena, this looks good. But we are missing the a few things:

a) The example provided by the task's description shows a `cpt` (a shorthand for Cascara Prop Types) object being destructured from the Cascara import. This suggests that all propTypes are part of that object. However, in this PR propTypes are being exported all by themselves.

b) This task applies to all Developer-facing components such as modules, Table, Form, Structures, As well as Form and Table actions (these are not components but have their own propTypes too), etc. This PR contains some of them but not all.

c) We also need to have our docs updated to let Developers know that we are exporting all prop types and how to use them.

d) Add a changeset

If you want, we can have a call to chip-out this bad boy together 👍

brianespinosa:
@rociobaena @lordpixel I think there is some confusion around what this ticket is about. This is not about us consistently exporting all of the type shapes from our components (though that is I think something we should be doing to help ourselves outside of Cascara). This ticket is about creating custom prop types for use when defining types on a component. Basically we want to extend what prop types are available for building types. I will expand on this in the ticket itself. I thought it was going to be clear what our use case was with the code example on the ticket... but when I look at it now, I can see why it is not clear. Sorry!

brianespinosa:
@rociobaena When there is a PR we put up that we are just getting feedback on, we can set the PR as a draft in Github. Please go ahead and do that with this one so we know it is not supposed to be merged yet.

rociobaena:
This implementation was not no the expected one. Declining this in favor of https://github.com/Espressive/cascara/pull/312 instead.

PR #311, opened 2021-11-23
feat(FDS-382): [Cascara] Stat component mvp

lordpixel:
## Resolves [FDS-382](https://espressive.atlassian.net/browse/FDS-382).

This creates the initial iteration of a Stats component MVP.

## New Component Checklist

### DIFF
--- a/packages/cascara/src/ui/Stat/Stat.js
+++ b/packages/cascara/src/ui/Stat/Stat.js
@@ -1,0 +1,62 @@
+import React, { useCallback } from 'react';
+import pt from 'prop-types';
+import classNames from 'classnames/bind';
+
+import { Clickable, Role } from 'reakit';
+import styles from './Stat.module.scss';
+
+const propTypes = {
+  /** Stats can have their own css class name */
+  className: pt.string,
+  /** Stats can be fuild */
+  fluid: pt.bool,
+  /** Stats can have a label */
+  label: pt.string,
+  /** Stats can be clickable */
+  onClick: pt.func,
+  /** Stats can have a sub text */
+  sub: pt.oneOfType([pt.number, pt.string]),
+  /** Stats can have a value */
+  value: pt.oneOfType([pt.number, pt.string]),
+};
+
+const cx = classNames.bind(styles);
+
+const WidgetStatsStat = ({ className, fluid, onClick, label, value, sub }) => {
+  // Instead of maintaining separate, competing styles for focus, we are setting focus on this clickable item on hover. This may be something we consider doing on other Clickable components with Reakit.
+  const handleFocus = useCallback(({ currentTarget, dispatchConfig }) => {
+    const { registrationName } = dispatchConfig;
+
+    if (registrationName === 'onMouseEnter') {
+      currentTarget.focus();
+    } else if (registrationName === 'onMouseLeave') {
+      currentTarget.blur();
+    }
+  }, []);
+
+  return (
+    <Role
+      as={onClick ? Clickable : 'div'}
+      className={cx('Stat', className, {
+        Clickable: Boolean(onClick),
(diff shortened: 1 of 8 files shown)


packages/cascara/src/ui/Stat/__globals.js
```diff
@@ -0,0 +1,16 @@
...
+  /** Stats can be fuild */
+  fluid: pt.bool,
+  /** Stats can have a label */
+  label: pt.string,
+  /** Stats can be clickable */
+  onClick: pt.func,
+  /** Stats can have a sub text */
+  sub: pt.oneOfType([pt.number, pt.string]),
+  /** Stats can have a value */
+  value: pt.oneOfType([pt.number, pt.string]),
```
brianespinosa, line 15: We have these prop types being defined in two places. Which is our source of truth?

brianespinosa (changes requested):
@lordpixel a few small changes. I think this is generally what we are looking for. Once those few things are done, this should be ready to come in too!

packages/cascara/src/ui/Stat/__globals.js
lordpixel: The one in the component. Props were initially put here but Docs doesn't support showing the props table that isn't part of the component file.

packages/cascara/src/ui/Stat/Stat.js
```diff
@@ -0,0 +1,62 @@
...
+  /** Stats can have a sub text */
+  sub: pt.oneOfType([pt.number, pt.string]),
+  /** Stats can have a value */
+  value: pt.oneOfType([pt.number, pt.string]),
+};
+
+const cx = classNames.bind(styles);
+
+const WidgetStatsStat = ({ className, fluid, onClick, label, value, sub }) => {
+  // Instead of maintaining separate, competing styles for focus, we are setting focus on this clickable item on hover. This may be something we consider doing on other Clickable components with Reakit.
```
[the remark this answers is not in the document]
lordpixel: ok

packages/cascara/src/ui/Stat/Stat.js
```diff
@@ -0,0 +1,62 @@
...
+    if (registrationName === 'onMouseEnter') {
+      currentTarget.focus();
+    } else if (registrationName === 'onMouseLeave') {
+      currentTarget.blur();
+    }
+  }, []);
+
+  return (
+    <Role
+      as={onClick ? Clickable : 'div'}
```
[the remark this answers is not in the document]
lordpixel: this makes sense, I applied the change, thanks 👍

lordpixel:
> @lordpixel a few small changes. I think this is generally what we are looking for. Once those few things are done, this should be ready to come in too!

@brianespinosa this is ready for 👀 , 👍

packages/cascara/src/ui/Stat/__globals.js
brianespinosa: @lordpixel if we are not using these, let's remove them then. Leaving them is going to confuse someone coming along who might see these and try to change them. If we are forced for now to use the propTypes inside the file, we should keep those instead of these.

I need to look at React Docgen to see if they have the updates yet that will allow us to compose external props.

brianespinosa (changes requested):
We need to only have one source for definition of propTypes so let's remove the one in __globals for now.

packages/cascara/src/ui/Stat/__globals.js
lordpixel: @brianespinosa it makes sense, I forgot to remove the declaration in the globals when I realized that those weren't picked up by the propsTable component. 👍

brianespinosa (approved):

PR #312, opened 2021-12-01
chore(FDS-376): [@espressive/prop-types] - Some official custom prop types

rociobaena:
Export custom prop types.
Hi everyone. This is the draft for the package prop-types. Take a look at the Button.js. I used it there and it is working.
Waiting for your feedback now :)

### Dependencies
New packages is added to create our custom prop list.

## PR Author Checklist


package.json
```diff
@@ -73,6 +74,7 @@
     "@bjeco/prettier-config": "0.1.6",
     "@changesets/cli": "2.17.0",
     "@espressive/eslint-config": "workspace:*",
+    "@espressive/prop-types": "workspace:*",
```
brianespinosa, line 77: This dependency is not used at the root of our monorepo. The only dependencies that should be declared at the root are typically those that are being used at build time. I see there are others here that should not be present as well like fluent-ui, which is probably confusing.

Go ahead and leave this out for now at the root. It should only be present in packages that use it. For instance... it should be included as a devDependency for `@espressive/cascara`

utils/prop-types/package.json
```diff
@@ -0,0 +1,21 @@
...
+  "author": "Rocio Baena rbaena@espressive.com",
+  "main": "src/index.js",
+  "files": [
+    "index.js",
+    "src"
+  ],
+  "scripts": {
+    "prepare": "echo 'No prepare scripts for this package'"
+  }
+}
```
brianespinosa, line 21: This package should have a `peerDependency` defined as `prop-types` as it will need `prop-types` in order to work. The reason we probably cannot just include `prop-types` is that some libraries also require `prop-types` as a `peerDepencency` and those packages would not know that our custom prop types already have it.

We _also_ need to include `prop-types` as a `devDepencency` here. Any node module that we use in this package needs to be declared as a dependency in the package.json file. We use `prop-types` on line 1 of src/prop-types.js

utils/prop-types/src/prop-types.js
```diff
@@ -0,0 +1,15 @@
...
+  header: pt.oneOfType([pt.arrayOf(pt.node), pt.node]),
+  icon: pt.shape({}),
+  list: pt.oneOfType([pt.arrayOf(pt.node), pt.node]),
+  main: pt.oneOfType([pt.arrayOf(pt.node), pt.node]),
+  nav: pt.oneOfType([pt.arrayOf(pt.node), pt.node]),
+  post: pt.oneOfType([pt.arrayOf(pt.node), pt.node]),
+  title: pt.oneOfType([pt.arrayOf(pt.node), pt.node]),
+};
+const extendedPropTypes = { ...pt, ...customPropTypes };
+export default extendedPropTypes;
```
brianespinosa, line 15: The name of our default export should always match the file name. Our file name here is `prop-types`

brianespinosa (changes requested):

utils/prop-types/package.json
rociobaena: @brianespinosa @lordpixel can you look at this change, please? I understand of the peerDependency, but I'm not sure in I established the devDepency correctly:
`  "devDependencies": {
    "prop-types": "15.7.2"
  },
  "peerDependencies": {
    "prop-types": "15.7.2"
  }`

utils/prop-types/src/prop-types.js
rociobaena: **prop-types** is not a valid name. I leave it as propTypes instead.

utils/prop-types/package.json
lordpixel: @rociobaena LGTM 👍

lordpixel (approved):
Other than @brianespinosa's comments, nothing else to add 👍

packages/cascara/src/atoms/Button/Button.js
```diff
@@ -1,5 +1,5 @@
 import classNames from 'classnames';
-import pt from 'prop-types';
+import { propTypes as pt } from '@espressive/prop-types';
```
brianespinosa, line 2: We should not be importing a sub member of the package. We should be importing the default. If we have to import a named sub member of the package, we need to clean up the way the package code is defined.

packages/cascara/src/atoms/Flex/Flex.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import classnames from 'classnames/bind';
-import pt from 'prop-types';
+import { propTypes as pt } from '@espressive/prop-types';
```
brianespinosa, line 3: Should be using the default export, not named.

packages/cascara/src/atoms/Flex/FlexItem.js
```diff
@@ -1,12 +1,12 @@
 import React, { cloneElement } from 'react';
-import pt from 'prop-types';
+import { propTypes as pt } from '@espressive/prop-types';
```
brianespinosa, line 2: Should be using the default export, not named.

brianespinosa (changes requested):
Everything else looks correct here, but we still need to add this new package to the Changesets config. And we need to resolve the use of named exports from the package.

I will look into both of the above and also see what is going on with Vercel so this can be merged this morning.

@lordpixel I think this would also be good to include as a supporting package with Cascara v1 so keep an eye out for this one coming in before we make our develop > main PR this afternoon.

brianespinosa:
Had to update the Jest config so it could resolve the new package in tests. Other special packages are not used during tests so we have not run into this until now. 

Also added a Changeset, which was needed but not included.

@lordpixel it appears everything is passing now, though there was just a new build that kicked off.

brianespinosa (approved):

PR #314, opened 2021-12-08
feat(FDS-331): [Cascasa] - add OLD suffix to Table, Form and Chat

lordpixel:
Resolves [FDS-331](https://espressive.atlassian.net/browse/FDS-331).

Add `OLD` suffix to `Table`, `Form` and `Chat`.

### Fixtures
Fixtures updated to include the suffix

### Snapshots
Snapshots updated due to component name changes


lordpixel:
@brianespinosa we still need to rename all these components in NGAs and Middleware, my plan was to use the @next release of Cascara but this PR didn't generate one. Apparently, it generates these releases only when merging to develop. I will continue anyways, we can update after merging this.

brianespinosa (changes requested):
@lordpixel please go back and take a look at the ticket for naming conventions on Chat.

.changeset/perfect-tools-notice.md
```diff
@@ -0,0 +1,6 @@
+---
+"docs": patch
+"@espressive/cascara": patch
```
brianespinosa, line 3: @lordpixel this is not a patch change

docs/src/lib/MDX_COMPONENTS.js
```diff
@@ -51,19 +51,19 @@ const cascaraComponents = {
   ActionStack: (props) => <ActionStack {...props} />,
   Admin: (props) => <Admin {...props} />,
   Button: (props) => <Button {...props} />,
-  Chat: (props) => (
+  ChatOld: (props) => (
     <ChatProvider>
-      <Chat {...props} />
+      <ChatOld {...props} />
```
brianespinosa, line 56: These still say ChatOld

packages/cascara/src/ui/ChatMessages/Chat.doc.mdx
```diff
@@ -16,7 +16,7 @@ This is an array of objects used to tell Chat what to display. The `type` prop m
 ---
 title: Chat Messages
 ---
-<Chat
+<ChatOld
```
brianespinosa, line 19: Still says ChatOld

@lordpixel I am encountering a lot of these in code that still say ChatOld. Instead of me going line by line to find all of these, can we please do a search to find all of them and resolve first? Just doing a search in the DOM on this page I see 18 instances.

brianespinosa (changes requested):
There are still a bunch of old instances of ChatOld. Before I continue reviewing, could we please do a search to make sure we do not have this string anywhere?

.changeset/perfect-tools-notice.md
lordpixel: Changed to minor, 👍

docs/src/lib/MDX_COMPONENTS.js
lordpixel: corrected, thanks for the catch!

packages/cascara/src/ui/ChatMessages/Chat.doc.mdx
lordpixel: done, thanks for the catch!

.changeset/perfect-tools-notice.md
brianespinosa: @lordpixel anything that requires an end user to have to change their code to use our updates is always going to be a major, breaking change. If I as an end user can use this version without changing my code at all, then we know the change is not major. Due to the fact we are renaming components and removing component exports, this is absolutely a breaking change.

@rociobaena @lordpixel we should be aware of these rules here if we are ever unsure what level of change we are making with patch, minor, major: https://semver.org/

.changeset/spotty-parrots-sniff.md
```diff
@@ -0,0 +1,6 @@
+---
+"docs": patch
+"@espressive/cascara": patch
+---
+
+feat(FDS-331): [Cascasa] - add OLD suffix to Table, Form and Chat
```
brianespinosa, line 6: We only need one Changeset for this.

packages/cascara/src/ui/ChatMessages/themes/barista/components/Button/buttonStyles.js
```diff
@@ -1,4 +1,4 @@
-import * as _ from 'lodash';
+import { isNil } from 'lodash';
```
brianespinosa, line 1: I am going to create us a separate task to remove `lodash` completely as a dependency so we can keep our code more functional in style. Not in scope for this, but just a heads up that we will be removing it from Cascara. I think this was code for theming copied over directly from FluentUI themes. This is probably the place we are using it most... and also why there were direct imports of `* as _` going on.

No need to change these now as this will introduce further risk and should be done separately. Mostly a heads up. Go ahead and mark this conversation as resolved once you have seen it.

packages/cascara/src/ui/FormOld/index.js
```diff
@@ -0,0 +1 @@
+export { default, propTypes as formPropTypes } from './FormOld';
```
brianespinosa, line 1:
```suggestion
export { default, propTypes as formOldPropTypes } from './FormOld';
```
This little one snuck by us! I don't think we are using those props anywhere in Cascara though... only externally.

packages/cascara/src/ui/TableOld/index.js
```diff
@@ -1,2 +1,2 @@
 export { TABLE_SHAPE as tablePropTypes } from './__propTypes';
```
brianespinosa, line 1:
```suggestion
export { TABLE_SHAPE as tableOldPropTypes } from './__propTypes';
```

packages/cascara/src/ui/index.js
```diff
@@ -3,17 +3,17 @@
 // Let's keep these sorted alphabetically based on their source directory.
 
 export { default as ActionStack } from './ActionStack';
-export { default as Chat } from './Chat';
+export { default as ChatMessages } from './ChatMessages';
 export { default as Columns } from './Columns';
 export { default as Dashboard } from './Dashboard';
-export { default as Form, formPropTypes } from './Form';
+export { default as FormOld, formPropTypes } from './FormOld';
```
brianespinosa, line 9:
```suggestion
export { default as FormOld, formOldPropTypes } from './FormOld';
```

packages/cascara/src/ui/index.js
```diff
@@ -3,17 +3,17 @@
...
 export { default as List } from './List';
 export {
   default as Pagination,
   paginationPropTypes,
   usePaginationState,
 } from './Pagination';
 export { default as Section } from './Section';
-export { default as Table, tablePropTypes } from './Table';
+export { default as TableOld, tablePropTypes } from './TableOld';
```
brianespinosa, line 17:
```suggestion
export { default as TableOld, tableOldPropTypes } from './TableOld';
```

brianespinosa (changes requested):
@lordpixel The only thing missed in the code that I can see is we did not rename the prop types we are exporting to also support the old suffix. We are not exporting prop types for ChatMessages (yet) so we are not impacted there.

We need to revise our double changeset, and make sure we set it to the correct semver type, but other than that, this big one looks ready to go too.

brianespinosa (approved):
@lordpixel I resolved the remaining, small items in this PR related to the prop types and changeset so we can get a new build. Everything else looked good, so if this new build passes, we can merge. Thank you for taking the time to move these big chunks of code!

.changeset/spotty-parrots-sniff.md
brianespinosa: I removed this.

PR #315, opened 2021-12-09
feat(FDS-313): [Cascara] - Update the existing color theme for Dashboard

rociobaena:
Update the existing color theme to remove/replace any colors that might incorrectly communicate meaning (range, risk, positive, negative)

image
image

## New Component Checklist


brianespinosa (approved):
TY @rociobaena 🙌🏽 

@CaffeinatedAllie I am sure will be very excited seeing this screenshot, too.

PR #317, opened 2021-12-14
fix(FDS-397): [Table] - fix broken styles

lordpixel:
Resolves [FDS-397](https://espressive.atlassian.net/browse/FDS-397).

Fix a bug that was caused by a bad renaming of css classes.

### Snapshots
Snapshots updated due to a change in css classes.

## PR Author Checklist


lordpixel:
@brianespinosa I think it makes more sense to just rename Table and keep all children as they are, do you agree? Also, does this PR requires a changeset?

brianespinosa (approved):

PR #318, opened 2021-12-14
chore(FDS-398): [Github] - update PR checklist, add conventional commits

lordpixel:
Resolves [FDS-398](https://espressive.atlassian.net/browse/FDS-398).

Update PR checklist to require conventional commits.

## PR Author Checklist

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,5 +1,7 @@
 ## PR Author Checklist
 
+- [ ] PR title adheres to [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#specification)
+- [ ] All commits in this PR adhere to [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#specification)
 - [ ] I have finished reviewing the contents of this PR for accuracy
 - [ ] If needed, I have included a [Changeset](https://github.com/changesets/changesets) and it follows [Semantic Versioning principles](https://semver.org/)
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,5 +1,7 @@
 ## PR Author Checklist
 
+- [ ] PR title adheres to [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#specification)
+- [ ] All commits in this PR adhere to [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#specification)
```
brianespinosa, line 4: I think the first item with PR title is good. The second checklist item (line 4) isn't something that can be fixed within the scope of the PR if the code has already been committed. I think in that scenario we need to be covering that by reviewing code and coaching the team onto the right path with the commit messages. The bigger concern is the title which will end up becoming the single, historical commit in the history.

I think we should still be following conventional commits in the other commits that get squashed, but probably not include it in the PR checklist.
```suggestion
```
```suggestion
- [ ] All commits in this PR adhere to [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#specification)
```

brianespinosa (changes requested):
Only one small change I think is needed here. I am going to go ahead and commit the change from the Github tools just so this comes in faster.

Thanks for improving this list, @lordpixel. This was a great, needed suggestion.

brianespinosa (approved):
Removed the line about all commits being conventional as we need to solve that by coaching/training.

PR #325, opened 2021-12-20
fix(FDS-265): [Dashboard] - do not render widget if no data is passed

lordpixel:
### Resolves [FDS-265](https://espressive.atlassian.net/browse/FDS-265).

Bubble widget blows up if data is not passed, for now, we are not rendering a widget instead.

## PR Author Checklist


packages/cascara/src/ui/Dashboard/Dashboard.js
```diff
@@ -50,32 +50,35 @@ const Dashboard = ({ as = 'div', config, ...rest }) => {
...
+                    <code key={key}>{key}</code>
+                    {widgetKeys.length > i + 1 && ', '}
+                  </>
+                ))}
+              </p>
+            </>
+          }
+        />
+      )
+    ) : null;
```
brianespinosa, line 81: @lordpixel we should make a distinction between undefined or empty data. Undefined data in Cascara should show as a loader. Empty data should show as an empty state.

Technically, the way the widgets are built for Dashboard, we should be able to go to the Widget component that has the title and shared widget logic and place a wrapper to render a loading state if there is no data to pass to any widget. 

I think returning `null` from any of our components is almost always going to be the wrong approach. From our system components, we probably never want to return nothing. If something should not render at all, that logic should exist out of Cascara. Otherwise, devs should be able to pass data as it comes in, and if that data happens to be undefined at the time of render, that means loading.

brianespinosa (changes requested):
@lordpixel I think we should not return null here. If we need to prevent rendering, that should happen outside of Cascara. If we are blowing up with undefined data, we should use the shared loading animation instead while we wait for data to come from FE.

brianespinosa (changes requested):
Coming back to review this, it appears we are still returning null. Cascara design patterns imply that for undefined data, we render a loading state, for empty object or array in data, we render an empty state. Returning null is not something we want to do in Cascara.

packages/cascara/src/ui/Dashboard/Dashboard.js
lordpixel: @brianespinosa I agee with you on the fact that rerturning null is not the correct approach. 

Checking `Widget` component, I can see we are already using css classes for undefined/empty data. Should that be ennough? please see [Widget.js](https://github.com/Espressive/cascara/blob/develop/packages/cascara/src/ui/Dashboard/widgets/Widget.js#L75)

brianespinosa: @lordpixel it seems that based on the updates to Widget.js that this bug report might not be valid anymore. That was what I was trying to ask on the ticket when I was wondering if this is still valid. Are we able to reproduce the reported problem anymore? If not, we can probably go ahead and decline this PR. If the problem still presents itself, I think there is a different issue going on.

lordpixel:
@brianespinosa closing this due to the issue not being repoducible. It seems that the fix in Widget is enough to prevent this bug from happenning.

PR #327, opened 2021-12-20
feat(FDS-396): [ChatMessages] - use ramda instead of lodash 

lordpixel:
### Resolves [FDS-396](https://espressive.atlassian.net/browse/FDS-396).

Use Ramda instead of Lodash. This should have a decent impact on our bundle size as well, since lodash is pretty big and is not benefitted by tree shaking. This will also keep us using functional programming styles.

## PR Author Checklist


rociobaena (approved):
pathOr looks interesting 🧐

brianespinosa (changes requested):
This is great. I can see that we are removing the use of lodash in the code. If this is done, the only thing missing is we need to remove lodash from our dependencies in all package.json or else we are still forcing people to install it and making it available for people to use.

In addition to removing it, I think it would be good for us to make a repo specific eslint rule to let people know that we have deprecated lodash in favor of ramda. This could be a separate change if you want, but we should at the very least remove lodash from package.json files.

lordpixel:
> This is great. I can see that we are removing the use of lodash in the code. If this is done, the only thing missing is we need to remove lodash from our dependencies in all package.json or else we are still forcing people to install it and making it available for people to use.
> 
> In addition to removing it, I think it would be good for us to make a repo specific eslint rule to let people know that we have deprecated lodash in favor of ramda. This could be a separate change if you want, but we should at the very least remove lodash from package.json files.

@brianespinosa `lodash` was not pesent in any package.json.

brianespinosa (approved):
Thanks for checking, @lordpixel Approving

PR #335, opened 2022-01-04
chore(FDS-427): [Cascara] - add @espressive/prop-types as peer-dep

lordpixel:
Closes [FDS-427](https://espressive.atlassian.net/browse/FDS-427).

### Dependencies
add `@espressive/prop-types` as peer-dep

## PR Author Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,13 +1,13 @@
 version: 2.1
 
 orbs:
-  node: circleci/node@4.7.0
-  slack: circleci/slack@4.4.4
+  node: circleci/node@5.0.0
+  slack: circleci/slack@4.5.1
 
 parameters:
   installed-key:
     type: string
-    default: cascara-MONOREPO_INSTALLED-{{ .Revision }}-14
+    default: cascara-MONOREPO_INSTALLED-{{ .Revision }}
 
 aliases:
   - &is_main
@@ -43,7 +43,9 @@ jobs:
     <<: *job_executor
     steps:
       - checkout
-      - node/install-packages:
+      - node/install-packages: # https://circleci.com/developer/orbs/orb/circleci/node#commands-install-packages
+          cache-only-lockfile: true
+          check-cache: always # yarn berry zero installs
           override-ci-command: yarn install --immutable
           pkg-manager: yarn
       - run:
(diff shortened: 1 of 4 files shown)


rociobaena (approved):

packages/cascara/package.json
```diff
@@ -65,6 +65,7 @@
   },
   "peerDependencies": {
     "@espressive/design-tokens": "*",
+    "@espressive/prop-types": "*",
```
brianespinosa, line 68: @lordpixel thank you for doing this.

I am noticing that the scoped dependency above that is existing, and this one, are not following the `workspace:` syntax. We should follow that syntax so we can be accurate with the versions expected, instead of using the wildcard, which can bite us in the butt later if people do not update their versions.

Could we make a change to this to change any of our Cascara monorepo peerDeps to make sure they are not using the wildcard style syntax and instead are using the yarn workspace syntax?

brianespinosa (changes requested):
Let's make sure all package dependencies, (dev, peer, full) are all using `"workspace:version"`

packages/cascara/package.json
lordpixel: That's how we did it:
image

I don't know why it was changed..

lordpixel: trying again..

lordpixel: That is how we did it, but I think `yarn` is chaging it...

#### Before install
image

#### After install
image

brianespinosa: Hmm. I will take a look at this too.

.circleci/config.yml
```diff
@@ -43,7 +43,10 @@ jobs:
     <<: *job_executor
     steps:
       - checkout
-      - node/install-packages:
+      - node/install-packages: # https://circleci.com/developer/orbs/orb/circleci/node#commands-install-packages
+          cache-only-lockfile: true
+          check-cache: always # yarn berry zero installs
+          cache-version: v2
```
brianespinosa, line 49: @lordpixel adding this fixed it. This is something new in the node orb, which I also updated. If we ever run into the scenario again where we see a cache that is failing to expand due to permissions, we can change this value in the cache version and it will cause the install to not find any previous cached installs and run from scratch.

brianespinosa (approved):

brianespinosa (approved):

PR #336, opened 2022-01-05
chore(FDS-399):[Node]: Update node engine version to 14.18.2

rociobaena:
Update node engine version to 14.18.2

### Dependencies
Update node version to the most recent version.

image
image

## PR Author Checklist

### DIFF
--- a/utils/eslint-config/package.json
+++ b/utils/eslint-config/package.json
@@ -39,7 +39,7 @@
     "eslint-plugin-testing-library": "^4.12.4"
   },
   "engines": {
-    "node": "^10.12.0 || >=12.0.0"
+    "node": "^10.12.0 || >=14.18.2"
   },
   "publishConfig": {
     "access": "public",
--- a/utils/eslint-config/package.json
+++ b/utils/eslint-config/package.json
@@ -38,9 +38,6 @@
     "eslint-plugin-react-hooks": "^4.2.0",
     "eslint-plugin-testing-library": "^4.12.4"
   },
-  "engines": {
-    "node": "^10.12.0 || >=14.18.2"
-  },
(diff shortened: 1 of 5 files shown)


utils/eslint-config/package.json
```diff
@@ -39,7 +39,7 @@
     "eslint-plugin-testing-library": "^4.12.4"
   },
   "engines": {
-    "node": "^10.12.0 || >=12.0.0"
+    "node": "^10.12.0 || >=14.18.2"
   },
```
brianespinosa, line 43: We can go ahead and remove this engines completely. We should not be building this package from within the package directory. And we should definitely not still allow node 10.

brianespinosa (changes requested):
One small change needed to the eslint package. Other than that, we just need to verify that when Vercel is building, we can see in the log output that they are now using node 14 as well. They _should_ be based on the package.json engines defined in the /docs package, but we want to check, just in case.

rociobaena:
> One small change needed to the eslint package. Other than that, we just need to verify that when Vercel is building, we can see in the log output that they are now using node 14 as well. They _should_ be based on the package.json engines defined in the /docs package, but we want to check, just in case.

Hi Brian.
This is from the Vercel settings:
image

utils/eslint-config/package.json
rociobaena: Engine removed from package.

brianespinosa (approved):
Thanks for doing this, @rociobaena!

_SO_ much easier than the Node 10 -> 12 update. :)

lordpixel (approved):
All good @rociobaena, I'd only ask to use conventional commits format for all commits. 👍

PR #341, opened 2022-01-06
chore(FDS-FDS-429):[Dependencies]: Remove semantic-ui-react as a dependency from Cascara

rociobaena:
Remove semantic-ui-react as a dependency from Cascara.

### Dependencies
Remove semantic-ui-react package since it is no longer used in our monorepo.

## PR Author Checklist


app-tests/cra-button/src/App.js
```diff
@@ -2,39 +2,12 @@ import React from 'react';
 import './App.css';
 import { Button } from '@espressive/cascara';
 
-import {
```
rociobaena, line 5: Hi @brianespinosa. After removing 'semantic-ui-react' from the root package.json, this file was not getting the reference. Considering that this is just for testing purposes and the component was to try the button, I kept that in place and remove the references for SUIR.

lordpixel: Hi team, I think removing the `semantic-ui-react` dependency is the way to go, since it is not used in Cascara anymore, it is not required in this demo app. @brianespinosa what do you think?

.changeset/slimy-avocados-thank.md
```diff
@@ -0,0 +1,6 @@
+---
+"docs": patch
```
lordpixel, line 2: @rociobaena let's remove `docs` from this changeset.

lordpixel (changes requested):

.changeset/slimy-avocados-thank.md
rociobaena: right!

rociobaena: Removed

app-tests/cra-button/src/App.js
brianespinosa: @rociobaena that page would not get the reference because in this PR SUIR was removed as a dependency of that test app. We do not need to remove SUIR from the test app as those are not part of the monorepo install process. These test apps are basically in isolation. We can keep the button here.

The point behind this specific test app was to see what our Cascara button looks like along the old SUIR button, and in an environment where the legacy CSS is also present. That is why we see both buttons. This is a case where we should leave it. Again, this is a separate test app and the dependency does not get installed unless someone is specifically in this test app. cc @lordpixel

brianespinosa (changes requested):
@rociobaena everything else in this PR looks good except for removing SUIR from the test app. We need that space for visual tests on new components we make to see how they look in a real environment alongside the old SUIR components. It does not impact dependencies of the code we ship.

.changeset/slimy-avocados-thank.md
brianespinosa: @rociobaena docs is still part of this changeset.

Please create a ticket for making a Danger rule that will check to make sure docs are not part of a Changeset, as I think this will help any new team members in the future, and we have seen this pop up a few times.

app-tests/cra-button/src/App.js
rociobaena: File recovered

packages/cascara/src/ui/FormOld/modules/ActionEdit/ActionEdit.js
```diff
@@ -95,7 +94,7 @@ const ActionEdit = ({
         onClick={handleCancel}
         type='button'
       >
-        <Icon name='delete' />
+        <i className='delete icon' />
```
brianespinosa, line 97: 👍🏽

brianespinosa (approved):
Thanks @rociobaena !

lordpixel (approved):

PR #342, opened 2022-01-07
chore(FDS-433):[Danger]: Danger should fail instead of sending a message. 

rociobaena:
Danger.js should fail instead of sending a message when the current branch is develop targeting to main.

## PR Author Checklist


lordpixel (approved):

.danger/index.ts
```diff
@@ -36,7 +36,7 @@ const isTargetBranchDevelopOrMain = ['main', 'develop'].includes(targetBranch);
 
 // validate if the current and target branch are specifically main or develop
 const isTargetBranchMain = targetBranch === 'main';
-const isCurrentDevelop = targetBranch === 'develop';
+const isCurrentDevelop = currentBranch === 'develop';
```
brianespinosa, line 39: @rociobaena this change is not what this PR is about. Take a look farther down the Dangerfile. There are different message types that Danger can throw. This ticket should make the build throw an `error()` when develop directly targets main. Right now it is only showing a `message()` which means that the build will still pass.

brianespinosa (changes requested):

.danger/index.ts
rociobaena: Hi @brianespinosa : There is a fail after that:
// fail danger if the current branch is develop intented to be merged to main
if (isDevelopComparedToMain) {
  fail(
    'Do not target main directly from develop. A separate release branch is needed'
  );
}

The message one is for check if the current and target branch is main or develop

brianespinosa (approved):

PR #343, opened 2022-01-08
chore(FDS-189):[Cascara]: Validate data-component usage on Cascara 

rociobaena:
Validate `data-component` usage on Cascara components

 Add the identifiable `data-component="ComponentName"`  to our components in Cascara.

### Snapshots
The new snapshots should reflect the data-component name.

## PR Author Checklist


lordpixel (approved):

lordpixel:
@rociobaena this looks good, we only need to update snapshots.

lordpixel (dismissed):

packages/cascara/src/ui/Columns/Columns.js
```diff
@@ -21,6 +21,8 @@ const propTypes = {
   style: pt.shape(pt.object),
 };
 
+const componentName = 'Columns';
+
```
brianespinosa, line 25: I don't see anywhere else in the component where we are using the `componentName` const. To me this feels like extra work for something that we should not be considering a configurable global. This data attribute should always be the name of the component itself. I think we should just put these strings directly on the `data-component='Columns'` attribute.

packages/cascara/src/ui/List/List.js
```diff
@@ -9,10 +9,17 @@ const propTypes = {
...
 
+const componentName = 'List';
 const List = ({ data, ...rest }) => {
   const compositeState = useCompositeState();
   return (
-    <Composite {...rest} className={styles._} {...compositeState} as='ul'>
+    <Composite
+      {...rest}
+      className={styles._}
+      {...compositeState}
```
brianespinosa, line 19:
```suggestion
      {...compositeState}
      className={styles._}
```
Let's move this spread value above className

brianespinosa (changes requested):
@rociobaena I am seeing that we are defining componentName as a const above the component and using that on the component below. Because are not using componentName anywhere else, this is extra abstraction that I do not think helps us right now. Let's go ahead and define this string directly on the data-component attribute.

rociobaena:
I close this PR due to a rename this branch to include a description, this includes the feedback provided here: https://github.com/Espressive/cascara/pull/345.

PR #346, opened 2022-01-12
chore(FDS-442): [prop-types] - Add CJS and ESM builds

lordpixel:
Resolves [FDS-442](https://espressive.atlassian.net/browse/FDS-442).

This updates the `rollup.config.js` file at the root of the monorepo, it allows for more controlled build process for packages other than `@espressive/cascara` that might not have scss or private files such as `@espressive/prop-types`.

### Dependencies
Changes to package.json file to add scripts and `walk-sync` dependency.

## PR Author Checklist


utils/prop-types/package.json
```diff
@@ -9,16 +9,25 @@
   },
   "license": "MIT",
   "author": "Rocio Baena <rocio.baena@espressive.com>",
+  "main": "dist/index.js",
   "module": "src/index.js",
```
brianespinosa, line 13: We should fix the `main` and `module` paths to be the same as they are in the Cascara package as these values are used directly in the rollup config to know where to output files.

If the Rollup config was run with these values in the package.json, you may have seen the files in src getting overwritten. I would even take a look at the index.js file there specifically as the Rollup config should be looking at this and it is essentially going to try to output esm directly into the src directory with this.

utils/babel-preset-espressive/package.json
```diff
@@ -13,7 +13,7 @@
     "src"
   ],
   "scripts": {
-    "prepare": "echo 'No prepare scripts for this package'"
+    "prepare": "echo 'No prepare scripts for babel-preset-espressive'"
```
brianespinosa, line 16: I like this. Thanks for making these strings more specific!

rollup.config.js
```diff
@@ -5,6 +5,7 @@ import { nodeResolve } from '@rollup/plugin-node-resolve';
 import postcss from 'rollup-plugin-postcss';
 import stringHash from 'string-hash';
 import cssUrl from 'postcss-url';
+import walkSync from 'walk-sync';
```
brianespinosa, line 8: I have a few concerns about us adding an extra dependency here. We should in theory just be able to use `fs.readdirsync` from node to do all of this. In fact... there are kind of examples of how this would work in our Docs, as we are using just basic node functions to crawl our repo directories and find the MDX files. Is there a strong reason we are adding another dependency here instead of just using node?

rollup.config.js
```diff
@@ -67,13 +68,38 @@ const external = (id) =>
...
+    walkSync(SOURCE_DIR, {
+      globs: ['private.js'],
+    }).length > 0;
+
   // Get the package.json file
   const pkgConfig = require(`${SOURCE_DIR}/package.json`);
+
   // Relative input location for Rollup to bundle from
-  const input = [`${SOURCE_DIR}/src/index.js`, `${SOURCE_DIR}/src/private.js`];
+  const input = [`${SOURCE_DIR}/src/index.js`];
```
brianespinosa, line 86: It looks like there are some pretty big changes in this file... but I am pretty sure the cause of our rollup config not working on everything else is this line here where we have the private entry.

I am pretty sure Rollup will just not use any of the other configured plugins if it does not encounter any files of those types. I suspect we might just be able to update this one thing and have a conditional input for the private exports in Cascara.

Were we seeing errors about the CSS plugins not being used? We could likely just check for private exports and call it a day.

utils/prop-types/package.json
lordpixel: working on it 👍

rollup.config.js
lordpixel: makes sense, working on it.

rollup.config.js
lordpixel: We saw some errors regarding rollup css plugins in `@espressive/prop-types` package. That's why we changed it not to use those that plugin if it's not needed.

rollup.config.js
lordpixel: Dependency removed.

utils/prop-types/package.json
lordpixel: This was updated as requested

rollup.config.js
lordpixel: ok, I tried reverting back to what it was and I didn't see the errors anymore, we are keeping all plugins.

lordpixel:
@brianespinosa I think the build should work now, but there are some issues with the cache (I think).

rollup.config.js
```diff
@@ -65,16 +66,32 @@ const external = (id) =>
...
+    input.push(PRIVATE_PATH);
+  }
 
-  // Shared Rollup plugins
+  // Shared Rollup plugins for code-only packages
   const rollupPlugins = [nodeResolve(), postcss(getPostCSSOptions()), json()];
 
+  // Gather info about the package
+  const IS_CJS = Boolean(pkgConfig?.main);
+  const IS_ESM = Boolean(pkgConfig?.module);
```
brianespinosa, line 89:
```suggestion
  const IS_ESM = pkgConfig?.type === 'module' || Boolean(pkgConfig?.module);
```
Technically we should also be checking the type first to determine if it is ESM. If there is a type then, the package should definitely be a module. Additionally, if `type === 'module'` then we should not have any CJS at all. So that would be the best way to check if ESM_ONLY. Any package that is only a module using type should not have CJS.

rousspb (approved):

rociobaena (dismissed):

.changeset/wild-kids-buy.md
```diff
@@ -0,0 +1,5 @@
+---
+"@espressive/prop-types": patch
+---
+
+chore(FDS-442): [Rollup] - Improve rollup config to handle code-only packages
```
brianespinosa, line 5:
```suggestion
chore(FDS-442): [@espressive/prop-types] - Added a main (cjs) and modules (esm) entry point to the package
```
If this change was only to the rollup config, we would not need a Changeset because we do not ship the Rollup config. However, this update to the config changes the prop-types package and the changeset should describe the impact of that change.

rollup.config.js
```diff
@@ -65,16 +66,32 @@ const external = (id) =>
...
-  const input = [`${SOURCE_DIR}/src/index.js`, `${SOURCE_DIR}/src/private.js`];
+  const input = [`${SOURCE_DIR}/src/index.js`];
+
+  // Not all packages have private exports
+  if (fs.existsSync(PRIVATE_PATH)) {
+    input.push(PRIVATE_PATH);
+  }
 
-  // Shared Rollup plugins
+  // Shared Rollup plugins for code-only packages
```
brianespinosa, line 84:
```suggestion
  // Shared Rollup plugins
```
This should go back to the previous comment. This change only has to do with detecting if there is a private input/output.

rollup.config.js
```diff
@@ -65,16 +66,32 @@ const external = (id) =>
...
   const rollupPlugins = [nodeResolve(), postcss(getPostCSSOptions()), json()];
 
+  // Gather info about the package
+  const IS_CJS = Boolean(pkgConfig?.main);
+  const IS_ESM = pkgConfig?.type === 'module' || Boolean(pkgConfig?.module);
+  const TYPE = pkgConfig?.type;
+
+  // check if only ESM is part of the output
+  const ESM_ONLY = IS_CJS && !IS_ESM && TYPE === 'module';
```
brianespinosa, line 93: This logic is still inconsistent. Will revise to align with the presence of "module" or not.

brianespinosa (changes requested):
Updating this PR with items to change so I can go in an get these last few things cleaned up and bring this in since Manu is out today. 😷

rollup.config.js
```diff
@@ -100,7 +117,8 @@ const getRollupConfig = ({ pwd, babelConfigFile }) => {
     external,
     input,
     output: {
-      dir: `${SOURCE_DIR}/${pkgConfig.main.replace('/index.js', '')}`,
```
brianespinosa, line 103: We do not want to hard code the dir here because someone might want to configure a different directory for these files.

rollup.config.js
```diff
@@ -67,10 +68,18 @@ const external = (id) =>
...
   const pkgConfig = require(`${SOURCE_DIR}/package.json`);
+
   // Relative input location for Rollup to bundle from
-  const input = [`${SOURCE_DIR}/src/index.js`, `${SOURCE_DIR}/src/private.js`];
+  const input = [`${SOURCE_DIR}/src/index.js`];
+
+  // If we have a private export, add it to our list of inputs
+  if (fs.existsSync(PRIVATE_PATH)) {
+    input.push(PRIVATE_PATH);
+  }
```
brianespinosa, line 82: After tests rolling back a few other things, this is really the only change that was needed. This rollup config should work with anything that has proper entry points defined now. The only edge case was us having a private export, which is now supported with this check. Everything else works correctly as it was (minus a discovery of an error about exports: auto in CJS)

brianespinosa (approved):

PR #351, opened 2022-01-13
chore(FDS-244):[Theme]: Color processing is defensive

rociobaena:
Make sure all of our theme color processing is 100% defensive no matter what

`useThemeCustomProperties()` is empty:
image

` useThemeCustomProperties('i am a string')` is string:
image

`useThemeCustomProperties({ something: 'i am not a color' })` is an object but with not color property
image

`useThemeCustomProperties(theme)` is well defined
image

## PR Author Checklist


lordpixel (dismissed):
@rociobaena I am approving this since it LGTM. The build is failing because we need to update the `rollup.config.js` file and added some dependencies to the packages. You can see  more details in [this PR](https://github.com/Espressive/cascara/pull/346).

packages/cascara/src/structures/hooks/useThemeCustomProperties.js
```diff
@@ -67,10 +67,13 @@ const addCustomProperties = (color) => {
 const removeCustomProperties = () =>
   document.getElementById(STYLE_IDENTIFIER).remove();
 
-const useThemeCustomProperties = ({ color }) => {
+const useThemeCustomProperties = ({ color } = {}) => {
```
brianespinosa, line 70:
```suggestion
const useThemeCustomProperties = ({ color }) => {
```
I _think_ we don't need to do this because if we do the check for color on line 73, and color is undefined, it should evaluate to false. I am pretty sure the change we need is just to check for `color` before we try to use it in the `addCustomProperties` function, right?

brianespinosa (changes requested):

packages/cascara/src/structures/hooks/useThemeCustomProperties.js
brianespinosa: @rociobaena I think this is the only thing we need to fix. Otherwise LGTM. Please @ me if this change gets made, or if this still fails without defining an empty object as default.

rociobaena: Yes, that's right @brianespinosa. Let me play with that just a little with that, and then I will update the PR

brianespinosa (approved):

PR #354, opened 2022-01-20
chore(FDS-436) docs Rule component

brianespinosa:
Adds a new Rule component to the docs. Also enables us to now write tests within the /docs directory, as previously they would not be recognized. Finally, this also resolves some issues in Columns that I discovered where we were not getting the correct stacking due to a CSS shorthand, and our class naming when coming from a file with the same class name as the file name, we were not using an underscore.

For this last issue, I will be creating a separate cleanup task so we can go through and make sure we are consistent on that, and set some guidelines for the styles and module class names.

### Snapshots

Two snapshots were updated due to root class name changes to underscores when the root class matched the file name. And there was one additional class name added to Columns which required a snapshot update.

## PR Author Checklist


packages/cascara/pages/30-structures-and-layouts.mdx
```diff
@@ -37,6 +37,13 @@ Examples of functional layouts:
 - Chat
 - List View User Directory
 
+<Rule
+  title='A Rule'
+  body='Do not do any bad things here or else we will cut you.'
+  doImg='https://placedog.net/500/500?id=1'
+  dontImg='https://placedog.net/500/500?id=2'
+/>
```
brianespinosa, line 45: This is going to get changed before the release when I take Allie through the docs on Thursday.

lordpixel (approved):
LGTM, 👍 

I am approving this PR since it all looks good to me. Only thing we are missing is to adhere to conventional commits, this is important just in case we do not squash-merge and keep the whole history.

It is true that squash-merge of PRs does not keep the whole commit history, but the more we do it, the best we adopt the process.

cc @rociobaena

PR #355, opened 2022-01-21
chore(FDS-438): Move Boundaries Colums Stat Title into atoms folder

lordpixel:
### Resolves [FDS-438](https://espressive.atlassian.net/browse/FDS-438).

image

### Snapshots
Snapshots changed because come components were moved to a new folder.

## PR Author Checklist


packages/cascara/src/atoms/Title/Title.js
```diff
@@ -1,8 +1,8 @@
 import React from 'react';
 import { Role } from 'reakit';
 
-import { Boundaries } from '../../system-components';
-import { Flex } from '../../atoms';
+import Boundaries from '../../atoms/Boundaries';
+import Flex from '../../atoms/Flex';
```
brianespinosa, line 5: @lordpixel are we importing directly versus a sub export to resolve circular dependencies? 

I am asking because in packages/cascara/src/modules/ModuleErrorBoundary/ModuleErrorBoundary.js below we are importing a sub member. If we need to import directly instead of the sub member import, let's be consistent about which style we use.

IMO if it is going to make us less prone to circular dependency errors. This might also mean we should rethink having an index.js in each section directory where we export all defaults so it becomes impossible to import from the directory.

Let me know if there is a reason for the difference in style here before we make some larger determinations.

packages/cascara/src/structures/AdminStructure/AdminStructure.js
```diff
@@ -4,7 +4,7 @@ import pt from 'prop-types';
 import { useStructureName, useThemeCustomProperties } from '../hooks';
 import styles from './AdminStructure.module.scss';
 import AdminStructureChildren from './AdminStructureChildren';
-import Boundaries from '../../system-components/Boundaries';
+import Boundaries from '../../atoms/Boundaries';
```
brianespinosa, line 7: Here's another one where we are importing the default...

packages/cascara/src/ui/ChatMessages/ChatAttachment.js
```diff
@@ -8,7 +8,7 @@ import {
   Ref,
 } from '@fluentui/react-northstar';
 import { FilesEmptyIcon } from '@fluentui/react-icons-northstar';
-import { Boundaries } from '../../system-components';
+import { Boundaries } from '../../atoms';
```
brianespinosa, line 11: And then one where we are importing a named export from the directory

packages/cascara/src/atoms/Title/Title.js
lordpixel: Yes @brianespinosa, we are resolving circular dependencies.

packages/cascara/src/structures/AdminStructure/AdminStructure.js
lordpixel: yup, same issue, circular dependencies.

packages/cascara/src/ui/ChatMessages/ChatAttachment.js
lordpixel: We should probably use the direct import, what do you think? @brianespinosa

brianespinosa: @lordpixel I agree then. Let's officially move to support only the direct import syntax.

I am going to approve this PR and have a task to remove all of the directory exports and have a singular export file with everything then. That will prevent anyone from doing anything but the direct export that we want from now on. Thank you for clarifying.

brianespinosa (approved):
LGTM. Enforcing direct exports can move to a separate task. Thanks @lordpixel

brianespinosa:
Added this task to track enforcing use of direct exports https://espressive.atlassian.net/browse/FDS-448

PR #358, opened 2022-01-24
chore(FDS-451): [Performance] Build output now exports to separate component files

brianespinosa:
This may need some testing in a production setup after merge to develop, but this appears safe.

Builds now do not output to only a couple monolithic JS files but instead output to separate chunks with matching dir structure as our source. This should make for better error logging for FE as the file that throws the error will now have a name that we can track to a specific Cascara file. Additionally, this is the first step for us to potentially also build a special package.json file later on that will allow importing a specific component only, which should allow for better tree shaking.

Small update also to the config so we check that the output is valid JS. This takes slightly longer, but is worth the extra check.

## PR Author Checklist


brianespinosa:
image

rociobaena (approved):

PR #361, opened 2022-01-25
chore(FDS-426) Resolve all YN0002 warnings

brianespinosa:
When doing an install, this should resolve all warnings with missing dependencies in the YN0002 range. These warnings were resolved in 3 ways:

- Add the correctly missing dependency where needed (there were only a few of these)
- Update some of the packages which had downstream missing dependencies which were resolved in later versions
- Define these missing dependencies in the yarnrc file

Note that when a dependency declared in the yarnrc file becomes redundant, Yarn will let us know that this is no longer needed and we can try removing it. This will usually be because a missing dependency was fixed. Some of these items will likely need to be PRs into the related OS package, or into the compat database, but the first step is resolving these here.

The only install warnings that should remain are YN0006 warnings about version mismatches. Those should get resolved when we update the eslint-config, which should be done separately.

### Dependencies

Some dependencies were updated in order to resolve some incorrect dependency resolutions in downstream packages

## PR Author Checklist


brianespinosa:
Note, when reviewing these PRs that have a lot of changed files in the cache for zero installs, we can use the filter tools to not show any of the .zip files. 
image

@lordpixel has a separate task for updating other dependencies. We will want to make sure that there are no new YN0002 warnings that come up when those updates happen. Moving forward, we will want to make yarn error on those types.

brianespinosa:
> Moving forward, we will want to make yarn error on those types.

I forgot there was a `logFilters` config possible in the yarnrc file which I just added. Now our installs will fail if any more `YN0002` issues show up.

rociobaena (approved):
Much better results.
image

PR #365, opened 2022-01-26
feature(FDS-383):[Dashboard]: Swap out to use the new Columns, Section, Stat components

rociobaena:
Swap out internal styles and components to use the new Columns, Section, Stat components internally from Cascara

### Snapshots
Stat was updating relying on the root class for `_` instead of `Stat` and adding a modifier class for clickable stats.

## PR Author Checklist


packages/cascara/src/components/Dashboard/Dashboard.module.scss
```diff
@@ -4,25 +4,17 @@ $widget-background: #f0f0f0;
 
 .Dashboard {
   display: block;
-  columns: 30rem;
+  // columns: 30rem; //verify we dont need this
```
rociobaena, line 7: Removing this, the main layout in the dashboard is positioned in a single column:
image

packages/cascara/src/components/Dashboard/Dashboard.module.scss
```diff
@@ -4,25 +4,17 @@ $widget-background: #f0f0f0;
...
   }
 
-  .Title {
-    display: block;
-    padding: 0.5em 0 0.25em;
-    margin: 0;
-  }
-
   .Data {
+    // we need this for Widget List
```
rociobaena, line 17: If we remove this, the rows lines disappear, that's why I keep them:
image

packages/cascara/src/components/Dashboard/widgets/WidgetStats.js
```diff
@@ -18,10 +17,12 @@ const WidgetStats = ({ data, ...rest }) => {
   const dataState = getDataState(data);
 
   return (
-    <Widget {...rest} {...dataState} className={styles.Stats} height='auto'>
```
rociobaena, line 21: The Columns component works so fine here. Just the appearance was altered:
image
image
Should we keep `styles.Stats` or should we work on the class of the Stat component?

brianespinosa: We can take a look. If I recall, there is a Stat specific override that is making the grey background go away on these that we will want to keep. But we can also possibly simplify the styles.

brianespinosa:
@rociobaena I am going to pull this branch today and take a look. Some of these I may need to poke around at to find out what is going on. Thank you for getting this started.

packages/cascara/src/components/Dashboard/Dashboard.module.scss
rociobaena: Commented addressed pairing with @brianespinosa.

packages/cascara/src/components/Dashboard/Dashboard.module.scss
rociobaena: This class is kept, no need to remove it.

lordpixel (approved):

PR #387, opened 2022-03-02
chore(FDS-489): [Monorepo] - bump prettier to v2.5.1

lordpixel:
This bumps prettier to its most recent version v2.5.1.

### Dependencies
`package.json` file updated because of the version bump.

## PR Author Checklist

### DIFF
--- a/ADR/prettier.md
+++ b/ADR/prettier.md
@@ -1,0 +1,47 @@
+# Prettier
+
+version: 2.5.1
+Date: 2021-12-02
+
+[diff](https://github.com/prettier/prettier/compare/2.5.0...2.5.1)
+
+#### Improve formatting for empty tuple types ([#11884](https://github.com/prettier/prettier/pull/11884) by [@sosukesuzuki](https://github.com/sosukesuzuki))
+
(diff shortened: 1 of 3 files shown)


ADR/prettier.md
```diff
@@ -0,0 +1,47 @@
+# Prettier
```
odiseo42, line 1: Is there any reason to keep this ADR? I see that the upgrade was successful without any implications. If there's nothing to discuss in the future I would say there's no need to have an ADR, or even delete the existing ADR if the decision or technical problem discussed in the document has already been solved.

odiseo42 (approved):
LGTM other than the existential question on the ADR doc

ADR/prettier.md
lordpixel: This is actually an error on my side, thanks for catching it. Let me add the AD record. 👍

lordpixel: Actually, I think it's important to keep this as a record of when was the last time we updated and if there were any important things to mention. What do you think, should we keep it?

PR #388, opened 2022-03-02
chore(FDS-485): [Monorepo] - bump husky to v4.3.0

lordpixel:
# Bump husky to a most recent version.

version: 4.3.0
Date: 2020-09-07

### Notes

We didn't go all the way to [v7.0.4](https://github.com/typicode/husky/releases/tag/v7.0.4) because in v5.0.0 they changed the license from MIT to The Parity Public License, we cannot use it for commercial purposes.
v4.3.0 seems to work just fine without us going and changing any config. We'll stay with that one.

### Features

- Add .cjs config file support #754

### Ecosystem updates

- none

## PR Author Checklist


ADR/husky.md
```diff
@@ -0,0 +1,17 @@
+# Husky
+
+version: 4.3.0
+Date: 2020-09-07
+
+### Notes
+
+We didn't go all the way to [v7.0.4](https://github.com/typicode/husky/releases/tag/v7.0.4) because in v5.0.0 they changed the license from MIT to The Parity Public License, we cannot use it for commercial purposes.
```
odiseo42, line 8: Thanks for keeping an eye on the licensing part.

What would be the long term implications? do we have a plan to replace husky?

odiseo42 (approved):
Looks so far to me. It made me nervous seeing licensing changes in Husky, I think we gotta generate a plan to move forward without it.

ADR/husky.md
lordpixel: They have been switching between license types, we just need to make sure we use one with MIT.

Regarding replacing husky: Husky is tied to lint-staged which in turn is linked to git, at the moment IDK of any library that can replace Husky. I'll do some explorations in the ecosystem. 👍

odiseo42: @lordpixel ok me understand what you're saying. Husky v7 is already listed as MIT https://github.com/typicode/husky/blob/v7.0.4/LICENSE —  why wouldn't we be allowed to upgrade to version 7? If they changed to another licensing type during 5.0 then I don't think that prevents us from making the jump to 7, or am I missing something?

PR #391, opened 2022-03-04
chore(FDS-498) [Monorepo] - update typescript

lordpixel:
# Typescript

Version: 4.6.2
Date: 2022-03-04

> [Release notes](https://github.com/microsoft/TypeScript/releases/tag/v4.6.2).

### Ecosystem updates

- none

### Snapshots
Snapshots updated due to aun update in dependencies.

### Dependencies
Dependencies updated due to bump in library versions.

## PR Author Checklist

### DIFF
--- a/ADR/typescript.md
+++ b/ADR/typescript.md
@@ -1,0 +1,10 @@
+# Typescript
+
+Version: 4.6.2
+Date: 2022-03-04
+
+> [Release notes](https://github.com/microsoft/TypeScript/releases/tag/v4.6.2).
+
+### Ecosystem updates
+
+- none
--- a/package.json
+++ b/package.json
@@ -126,7 +126,7 @@
     "rollup": "2.67.3",
     "rollup-plugin-postcss": "4.0.2",
     "sass": "1.34.1",
-    "typescript": "4.3.2",
+    "typescript": "4.6.2",
     "vercel": "22.0.1",
     "webpack": "4.44.1"
   },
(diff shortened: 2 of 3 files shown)


ADR/rollup.md
```diff
@@ -0,0 +1,33 @@
+# Rollup
+
+version: 2.67.3
+Date: 2022-02-02
+
+### Features
+
+- Improve side effect detection when using Array.prototype.groupBy/groupByToMap (#4360)
```
odiseo42, line 8: I think we are using the ADR here to keep a copy of the changelog of a library itself — I would prefer if we keep our ADRs only for those changes, discussions or problems that affect our internal code. 
For instance if there's a library update that introduces API changes that make our code incompatible, that's the kind of information we would want to keep in the ADR.

docs/src/lib/MDX_COMPONENTS.js
```diff
@@ -10,17 +10,17 @@ import {
   AreaPlaceholder,
   Boundaries,
   Button,
-  ChatMessages,
```
odiseo42, line 13: I'm confused, why is this renaming of components happening in this PR?

odiseo42 (commented):
Two pending comments form me in this PR

docs/src/lib/MDX_COMPONENTS.js
lordpixel: This is not correct. I am investigating, I did not touch this file.

lordpixel: The PR was pointing to `main`, which is incorrect. Branch revisited and correctly set. 👍

PR #411, opened 2022-03-16
feat(FDS-518): [Table] - Record sorting PoC

lordpixel:
This is a PoC for `record sorting` in Cascara Table.

## How to test
- `yarn install`
- `yarn cosmos`
- Navigate to Table section

### All Attributes short-hand flag

In this scenario, ALL attributes are sortable:
image
> Note the `sortable` prop is passed as a `Boolean`.

### Arbitrary Attribute List Sort (Array)

In this scenario, `eid` and `name` attributes are sortable:
image
> Note the `sortable` prop is passed as `Array[String]`.

### Single Attribute Sort

In this scenario, only one attribute `eid` is sortable:
image

> Note the `sortable` prop is passed as `String`.

### Fixtures

A new Cosmos Fixture was created to support unit tests.

## PR Author Checklist


packages/cascara/src/components/Table/state/sortingReducer.js
```diff
@@ -0,0 +1,44 @@
+const SORT = 'SORT';
+const SORT_ORDER = {
```
odiseo42, line 2: Nice little mini reducer

odiseo42 (approved):
I think this all looks fine! You should demo this (in Cosmos) to see if there's any additional request before we do the implementation on app_web

packages/cascara/src/components/Table/TableBase.js
```diff
@@ -129,7 +164,24 @@ const TableBase = ({
     ? inferUniqueID(Object.keys(data[0]))
     : undefined;
 
-  // // FDS-142: new action props
+  // FDS-518: sort table records
+  const sortedData = useMemo(() => {
+    if (!sortState?.attribute || sortState?.order === SORT_ORDER.UNSORTED) {
```
andresaraujo, line 169: (Optional) `attribute` and `order` property names always exist in the state, you could drop the `?.` null safe operator

packages/cascara/src/components/Table/TableBase.js
```diff
@@ -129,7 +164,24 @@ const TableBase = ({
     ? inferUniqueID(Object.keys(data[0]))
```
andresaraujo, line 164: (Question) Wondering if you need to do this with sortedData instead

andresaraujo (commented):
Looking good.

What would happen if you select a row and then sort the column? does it keep the correct row selected?

packages/cascara/src/components/Table/atoms/SortableHeader.js
```diff
@@ -0,0 +1,54 @@
...
+    [styles.SortArrowDown]:
+      isSortAttribute && sortState?.order === SORT_ORDER.DESCENDING,
+    [styles.SortArrowUp]:
+      isSortAttribute && sortState?.order === SORT_ORDER.ASCENDING,
+    [styles.SortInactive]:
+      !isSortAttribute || sortState?.order === SORT_ORDER.UNSORTED,
+  });
+
+  /** Handle SVG element click */
+  const handleSVGClick = useCallback(() => {
```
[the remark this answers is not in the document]
lordpixel: You are right @andresaraujo, a previous implementation used SVG but it didn't work as we wanted so we removed it. We didn't rename the click handler. Thanks for the catch, renaming 👍

packages/cascara/src/components/Table/TableBase.js
lordpixel: Thanks for catching this, unique ID inference works with the `data` shape, which isn't impacted by sorting. That's why we didn't change it. But I agree we could use `sortedData` for consistency, changing 👍

packages/cascara/src/components/Table/TableBase.js
lordpixel: This is a remanent of the time when we were building the reducer. It's ok to remove it now. Tanks for the catch!

andresaraujo (approved):

